PR #3, opened 2024-06-27
The sources produce SEDs instead of point observations

jeremykubica:
Return a T x N matrix of SED observations from each source where T is the number of time stamps and N is the number of wavelengths we are sampling.


hombit (approved):
Loog good!

OliviaLynn (approved):
Looks good!

PR #13, opened 2024-06-28
Create better resampling

jeremykubica:
Currently TDAstro requires the creation of a new `PhysicalModel`s anytime we resample the parameters. This might be expensive. This PR adds a slightly more flexible model where the `PhysicalModel` can pull attributes samples from functions. It is still a bit awkward because the function parameters need different names.

A next step will be to extend this to take in a DAG of sampling effects and do the sampling based on that DAG.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -1,5 +1,23 @@
+import types
+from enum import Enum
+
+
+class ParameterSource(Enum):
+    """ParameterSource specifies where a PhysicalModel should get the value
+    for a given parameter: a constant value, a function, or from the host.
+    """
+
+    # TODO support something like TensorFlow probability to provide a DAG
+    # of sampling dependencies.
+    CONSTANT = 1
+    FUNCTION = 2
+    HOST = 3
+
+
 class PhysicalModel:
-    """A physical model of a source of flux.
+    """A physical model of a source of flux. Physical models can have fixed attributes
+    (where you need to create a new model to change them) and settable attributes that
+    can be passed functions or constants.
 
     Attributes
     ----------
@@ -11,6 +29,9 @@ class PhysicalModel:
         The object's declination (in degrees)
     distance : `float`
         The object's distance (in au)
+    setters : `dict` or `tuple`
+        A dictionary to information about the setters for the parameters in the form:
+        (ParameterSource, value).
     effects : `list`
         A list of effects to apply to an observations.
     """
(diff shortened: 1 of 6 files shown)


src/tdastro/base_models.py
```diff
@@ -32,20 +53,85 @@ def __init__(self, host=None, ra=None, dec=None, distance=None, **kwargs):
...
-        if dec is None and host is not None:
-            self.dec = host.dec
-        if distance is None and host is not None:
-            self.distance = host.distance
+        Raises
+        ------
+        Raise a ``KeyError`` if there is a parameter collision.
+        Raise a ``ValueError`` if the parameter is required, but set to None.
+        """
+        if hasattr(self, name):
```
OliviaLynn, line 90:
```suggestion
        if hasattr(self, name) and getattr(self, name) is not None:
```
I wonder about adding this in for cases where we've set the attribute as None, but would like to override this (especially for attrs like `ra`, `dec`, and `distance`, that default to None in the base model).

As is, the following will raise a KeyError:
```
model = StepSource(
        brightness=_sample_brightness,
        t_start=0.0,
        t_end=_sample_end,
        magnitude=100.0,
        duration=5.0,
    )
    model.add_parameter('ra', 0.0)
```

However, I do not know how much we would want to encourage post-hoc setting of fundamental attributes like this.

OliviaLynn (approved):
Looks good! Just the one question

src/tdastro/base_models.py
```diff
@@ -32,20 +53,85 @@ def __init__(self, host=None, ra=None, dec=None, distance=None, **kwargs):
...
+        """Sample the model's underlying parameters if they are provided by a function.
 
-        self.effects = []
+        Parameters
+        ----------
+        **kwargs : `dict`, optional
+           All the keyword arguments, including the values needed to sample parameters.
+        """
+        if self.host is not None:
+            self.host.sample_parameters(**kwargs)
```
mi-dai, line 121: I'm a little confused why we need a separate recipe for host and can't just do 
```
host = PhysicalModel()
host_parameters = host.sample_parameters()
```

jeremykubica: My thinking is that when we resampled the `PhysicalModel`'s parameters, we would always want to make sure to resampled the host's parameters as well. However if we'd prefer to separate the two, then we could use the approach you suggest. What would be the most useful from the user perspective?

jeremykubica: We can definitely change the name. The basic idea behind special casing host is to allow: 1) the source to read information that isn't provided in the source (RA, dec, etc.) and 2) provide a name for sources to refer back to their host. We could rename it to something like "background" or get rid of the special casing altogether and just have it be an attribute.

PR #14, opened 2024-06-28
Provide a mechanism for collecting needed citations

jeremykubica:
Allow the user to automatically assemble needed citations based on the `PhysicalModel`s and `EffectModel`s that they use.


jeremykubica:
Closing for now. There is probably a better general purpose tool (or opportunity for one).

PR #16, opened 2024-07-02
Add redshift effect

### LINKED ISSUE #4: Add redshift effect
Add the ability to transform the SED matrix given a distance.

OliviaLynn:
## Change Description

Closes #4

### Redshift conversions
1. Converts desired `observation_times`, `observation_wavelengths` to `emitted_times`, `emitted_wavelengths`. 
1. We then feed these `emitted_times`, `emitted_wavelengths` values to a given PhysicalModel to get the `emitted_flux`. 
1. Finally, we convert the `emitted_flux` to `observed_flux` to get the flux densities as they would appear on Earth.

Conversions used are:
    - emitted_wavelength = observed_wavelength / (1 + redshift)
    - emitted_time = observed_time / (1 + redshift)
    - observed_flux = emitted_flux / (1 + redshift)

### Pre-effects
Note that this PR also creates the concept of "pre-effect," which we apply to times, wavelengths in the PhysicalModel's evaluate, just before we call `flux_density = self._evaluate(times, wavelengths, **kwargs)`. 

Our initial thoughts were that the only other effect that might implement a pre_effect method would be some kind of lensing (I think Mi said strong lensing?); if we do add another such pre_effect, we should consider the ordering and other factors that could be involved in them playing nicely together.


src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+            The wavelengths at which the observation is made.
+        **kwargs : `dict`, optional
+           Any additional keyword arguments.
+
+        Returns
+        -------
+        float
+            The adjusted flux density at the observed wavelength.
+        """
+        return (observed_times / (1 + self.pz), observed_wavelengths / (1 + self.pz))
```
hombit, line 54: I think we want to make it a bit more complicated with time. Our "time" is actually a "date", so we cannot multiply it - the result would have little sense. Also, it is not convenient to define time in the model to be "deredshifted" date, because than we cannot generate reference epoch (for example peak date or event start date) without knowing a redshift.

I think the only solution is to ask the model to define the "epoch" (reference time) which would be the time (date) in observer frame, and use it here. So all the times in the rest frame would be relative to this reference time:
```python
rest_times = (observed_times - epoch) / (1 + self.pz) + epoch
```

OliviaLynn: Notes from convo:

It should be fine to assume every PhysicalModel to have an epoch. (This is straightforward for, say, a supernova with a peak; for a variable star, Kostya says one would use one of the peaks. For say, an AGN, one might just define the epoch as the beginning of the survey)

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+    - emitted_time = observed_time / (1 + redshift)
+    - observed_flux = emitted_flux / (1 + redshift)
+    """
+
+    def __init__(self, pz, **kwargs):
+        """Create a Redshift effect model.
+
+        Parameters
+        ----------
+        pz : `float`
```
hombit, line 28: Minor comment: I think `pz` means photometric `z`, it would be better to use `z` or `redshift` here. I personally prefer longer variable names, so I vote for `redshift`

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Redshift(EffectModel):
+    """A redshift effect model.
+
+    This contains a "pre-effect" method, which is used to calculate the emitted wavelengths/times
+    needed to give us the observed wavelengths and times given the redshift.
+
+    Attributes
+    ----------
+    pz : `float`
```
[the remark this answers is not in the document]
OliviaLynn: Oops, I'm so used to pz for photometric redshift. I'll change this to redshift!

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+        super().__init__(**kwargs)
+        self.redshift = redshift
+
+    def pre_effect(self, observed_times, observed_wavelengths, **kwargs):
+        """Calculate the emitted wavelengths/times needed to give us the observed wavelengths
+        and times given the redshift.
+
+        Parameters
+        ----------
+        observed_times : float
```
jeremykubica, line 42: This should be a numpy.ndarray so that we can vectorize over all times and wavelengths. I don't think the code needs any changes, just the comment to:

`observed_times : float or numpy.ndarray`

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+            The times at which the observation is made.
+        observed_wavelengths : float
```
jeremykubica, line 44: This should be a numpy.ndarray so that we can vectorize over all times and wavelengths. I don't think the code needs any changes, just the comment to:

`observed_wavelengths : float or numpy.ndarray`

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,76 @@
...
+            The wavelengths at which the observation is made.
+        **kwargs : `dict`, optional
+           Any additional keyword arguments.
+
+        Returns
+        -------
+        float
```
jeremykubica, line 51: This doesn't match what is in the code. I think this should be a `tuple of (numpy.ndarray, numpy.ndarray)` and the comment should say something about the modified times and wavelengths.

jeremykubica (commented):
Looks really good. I will let the scientists do the actual review to check the correctness of the computations. I added a few small comments.

OliviaLynn:
Questions for tomorrow:
1. State of "required parameters"/is it ok to have redshift as a required param
2. Would like to go over pre-effect docstring, feels wordy and I don't know standard language in this area
3. Would like us to further discuss epochs/how we line up timing as part of the shift-and-scale adjustment that is our redshift calculation

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, redshift=None, t0=None, **kwargs):
+        """Create a Redshift effect model.
+
+        Parameters
+        ----------
+        redshift : `float`
+            The redshift.
+        t0 : `float`
+            The epoch of the peak or the zero phase, date.
```
OliviaLynn, line 27: WORDING (1/3)

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(**kwargs)
+        self.required_parameters = ["redshift", "t0"]
+        self.add_parameter("redshift", redshift, required=True, **kwargs)
+        self.add_parameter("t0", t0, required=True, **kwargs)
+
+    def __str__(self) -> str:
+        """Return a string representation of the Redshift effect model."""
+        return f"RedshiftEffect(redshift={self.redshift})"
+
+    def pre_effect(
```
OliviaLynn, line 41: WORDING (2/3) I use "observed" and "emitted" throughout the pre-effect code. Should I use something closer to "observation frame" and "rest frame"? (Is rest frame even applicable here?)

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+        observed_times : numpy.ndarray
+            The times at which the observation is made.
+        observed_wavelengths : numpy.ndarray
+            The wavelengths at which the observation is made.
+        **kwargs : `dict`, optional
+           Any additional keyword arguments.
+
+        Returns
+        -------
+        tuple of (numpy.ndarray, numpy.ndarray)
```
OliviaLynn, line 58: WORDING (3/3): Someone tied to previous. I'm not sure how to word what's going on here in a way that's readable, but still provides all the information needed.

OliviaLynn:
# Update July 10

## Changes:
1. **Main**: Time calculations now performed with respect to epoch aka `t0`
1. Converted the EffectModel's `required_parameters` from a method to an attribute
1. Uploaded a [redshift-demo notebook](https://github.com/lincc-frameworks/tdastro-sandbox/blob/orl/add-redshift-demo/redshift-demo.ipynb) to a branch in tdastro-sandbox

## Questions:
1. **Have some outstanding questions about nomenclature/general wording** - especially in 3 places, in links just below, but also highlighted in the PR with a comment per line-of-interest
  a. [L27:](https://github.com/lincc-frameworks/tdastro/pull/16#discussion_r1672870999) General wording; also, are we picking t0 or epoch?
  a. [L41:](https://github.com/lincc-frameworks/tdastro/pull/16/files#r1672875527) I use "observed" and "emitted" throughout the pre-effect code. Should I use something closer to "observation frame" and "rest frame"? (Is rest frame even applicable here?)
  a. [L58:](https://github.com/lincc-frameworks/tdastro/pull/16/files#r1672876769) Someone tied to previous. I'm not sure how to word what's going on here in a way that's readable but still provides all the information needed.

src/tdastro/base_models.py
```diff
@@ -158,18 +158,21 @@ class PhysicalModel(ParameterizedModel):
...
     distance : `float`
         The object's distance (in pc)
+    redshift : `float`
+        The object's redshift.
     effects : `list`
         A list of effects to apply to an observations.
     """
 
-    def __init__(self, ra=None, dec=None, distance=None, **kwargs):
+    def __init__(self, ra=None, dec=None, distance=None, redshift=None, **kwargs):
```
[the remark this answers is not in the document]
hombit: We need a cosmological model to derive it. For me it sounds reasonable to incorporate cosmological model into the `RedshiftEffect`, but maybe there is a cleverer way to decouple these two things

src/tdastro/effects/redshift.py
hombit: ... for periodic variable star.

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+        t0 : `float`
+            The epoch of the peak or the zero phase, date.
+            # TODO WORDING (1/3) -> Both how this is written, and to check, are we picking t0 or epoch?
+        **kwargs : `dict`, optional
+           Any additional keyword arguments.
+        """
+        super().__init__(**kwargs)
+        self.required_parameters = ["redshift", "t0"]
+        self.add_parameter("redshift", redshift, required=True, **kwargs)
+        self.add_parameter("t0", t0, required=True, **kwargs)
```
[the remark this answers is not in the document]
hombit: I believe that `redshift` and `t0` are physical model's parameters, not effect's parameters

src/tdastro/effects/redshift.py
hombit: "Calculate observer-frame times and wavelength"?

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+        tuple of (numpy.ndarray, numpy.ndarray)
+            The emission-frame times and wavelengths needed to generate the emission-frame flux densities,
+            which will then be redshifted to observation-frame flux densities at the observation-frame
+            times and wavelengths. # TODO WORDING (3/3)
+        """
+        observed_times_rel_to_t0 = observed_times - self.t0
+        emitted_times_rel_to_t0 = observed_times_rel_to_t0 / (1 + self.redshift)
+        emitted_times = emitted_times_rel_to_t0 + self.t0
+        emitted_wavelengths = observed_wavelengths / (1 + self.redshift)
+        return (emitted_times, emitted_wavelengths)
```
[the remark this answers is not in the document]
hombit: Since we'd like to have a uniform interface, I'd vote to all models provide the same `evaluate(times, wavelengths)`, where `times` are not relative to t0. Since model knows its t0, it is not a big deal to use it to get relative times, while computational overhead is negligible (and potentially zero with JIT)

tests/tdastro/effects/test_redshift.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+from tdastro.effects.redshift import Redshift
+from tdastro.sources.step_source import StepSource
+
+
+def get_no_effect_and_redshifted_values(times, wavelengths, t0, t1, brightness, redshift) -> tuple:
+    """Get the values for a source with no effects and a redshifted source."""
+    model_no_effects = StepSource(brightness=brightness, t0=t0, t1=t1)
+    model_redshift = StepSource(brightness=brightness, t0=t0, t1=t1, redshift=redshift)
+    model_redshift.add_effect(Redshift(redshift=model_redshift, t0=model_redshift))
```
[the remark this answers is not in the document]
hombit: I agree with Jeremy, we shouldn't pass redshifts and t0s twice to both classes, even implicitly. We might see bad bugs with these approach when parameters are random number generators...

tests/tdastro/effects/test_redshift.py
hombit, line 1: It would be nice to have a test checking that wavelengths are redshifted.

- We can make an astrophysical test with black-body model, because we know an analytical answer for bolometric intensity: it is ~(1+z)^4.
- Or we can just use a simple spectral model, e.g. F_nu ~ wavelength, and check that F_nu changes in a way we expect

src/tdastro/effects/redshift.py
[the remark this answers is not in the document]
mi-dai: OK. I'm fine with using `times` instead of relative times. I would still like to see if this is convenient enough for users after implementing some models.

tests/tdastro/effects/test_redshift.py
OliviaLynn: Created #36

src/tdastro/effects/redshift.py
```diff
@@ -0,0 +1,89 @@
...
+            distance.
+        **kwargs : `dict`, optional
+            Any additional keyword arguments.
+
+        Returns
+        -------
+        flux_density : `numpy.ndarray`
+            The redshifted results.
+        """
+        return flux_density / (1 + self.redshift)
```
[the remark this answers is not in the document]
jeremykubica: For this and my comments throughout: Olivia and I chatted about it today. If we make effects behave like other nodes in the graph, they will have consistent sampling behavior. In that case we will want to have the effect take in the `PhysicalModel`'s parameters. 

The pros of this approach is that 1) it keeps everything in the same graph formulation, 2) keeps the same sampling counter throughout the entire graph, 3) doesn't require a specialized interface for effects (the `required_parameters` list), and 4) allows for stateful effects.

The cons are: 1) as @hombit) notes there is a chance this could be misspecified during model creation, 2) it requires creating a unique effect for each physical model.

Let's go with the current approach for now. If this causes too much confusion, we can change back to stateless effects.

jeremykubica (approved):

PR #18, opened 2024-07-08
Allow DAG construction from models

jeremykubica:
Create a more flexible sampling model that is closer to the DAG. Instead of being able to sample parameters from three sources (a constant, a function, or a host's attribute) allow the following sources:
- A constant value
- A function
- The attribute of any other model (including a host, statistical distribution, or something else)
- A function from any other model (including a host, statistical distribution, or something else)

This removes the use of a pre-specified `host` attribute. It also creates a generalized `ParameterizedModel` that we can use for populations or even effects. The various `ParameterizedModel`s can form a DAG.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -4,70 +4,53 @@ from enum import Enum
 
 class ParameterSource(Enum):
     """ParameterSource specifies where a PhysicalModel should get the value
-    for a given parameter: a constant value, a function, or from the host.
+    for a given parameter: a constant value, a function, or from another
+    parameterized model.
     """
 
-    # TODO support something like TensorFlow probability to provide a DAG
-    # of sampling dependencies.
     CONSTANT = 1
     FUNCTION = 2
-    HOST = 3
+    MODEL_ATTRIBUTE = 3
+    MODEL_METHOD = 4
 
 
-class PhysicalModel:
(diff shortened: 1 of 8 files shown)


src/tdastro/base_models.py
```diff
@@ -95,30 +78,49 @@ def add_parameter(self, name, value=None, required=False, **kwargs):
             value = kwargs[name]
         if value is not None:
             if isinstance(value, types.FunctionType):
+                # Case #1: If we are getting from a static function, sample it.
```
OliviaLynn, line 81:
```suggestion
                # Case 1: If we are getting from a static function, sample it.
```
Tiniest tiniest nit, but while it's on the mind

OliviaLynn (approved):
This looks great!

src/tdastro/base_models.py
```diff
@@ -127,12 +129,70 @@ def sample_parameters(self, **kwargs):
...
+    can be passed functions or constants.
+
+    Attributes
+    ----------
+    ra : `float`
+        The object's right ascension (in degrees)
+    dec : `float`
+        The object's declination (in degrees)
+    distance : `float`
+        The object's distance (in au)
```
hombit, line 165: in cm? in pc?

jeremykubica: I changed to PC. We can change later if we want other units.

PR #20, opened 2024-07-09
Expand the DAG resampling

jeremykubica:
Two changes to improve how the parameter sampling works:
- Parameters are sampled in the order in which they are added (by changing the dictionary to a list). This allows for dependencies among the parameters.
- An `EffectModel` can have and resample functional parameters.


OliviaLynn (approved):
This looks great!

PR #22, opened 2024-07-09
Move __init__ comments into class comments

### LINKED ISSUE #15: Move constructor documentation to class-level docstring
Usually, the description of `__init__`/`__new__` arguments leaves under class-level docstring, not in these methods documentation. The reason is that users usually don't look into these methods docs, but into class-level docs only.

See [PEP 257](https://peps.python.org/pep-0257/):
> The class constructor should be documented in the docstring for its `__init__` method. Individual methods should be documented by their own docstring.

jeremykubica:
Closes #15 

Most of the `__init__()` level comments were duplicates anyway.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -30,13 +30,6 @@ class ParameterizedModel:
     """
 
     def __init__(self, **kwargs):
-        """Create a ParameterizedModel object.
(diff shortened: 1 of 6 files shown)


hombit (approved):
Looks good!

PR #25, opened 2024-07-10
Add a function to read in files in grid format

jeremykubica:
As a precursor to being able to load SALT2 and SALT3 models, we need to be able to read in the input files. This PR adds functionality to read in files in the same format the sncosmo uses (although uses astropy tables to read them instead of manually reading the files).


hombit (commented):
It looks good to me, but I'm not sure how general these tools should be and what scope should they have. If we are focused to load sncosmo-like data only, I'd ask @mi-dai to review it

mi-dai (approved):
Looks good to me if this is only for reading SALT2/3 model files or other files with the same format.

PR #26, opened 2024-07-11
Wrap sncosmo models

jeremykubica:
This PR does two things:
1) Wrap all the sncosmo models in a `PhysicalModel` that can be used in our flow. This implementation will be a slower due to levels of indirection, but will provide access to the library for anything we haven't implemented natively.
2) Add the ability to change the parameter setting for a `PhysicalModel`. This is needed to provide compatibility with sncosmo's `set()` function.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ dependencies = [
     "astropy",
     "numpy",
     "scipy",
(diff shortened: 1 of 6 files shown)


jeremykubica:
Closing to break this into two. #27 is the first half of this.

PR #27, opened 2024-07-11
Allow users to change parameter settings

jeremykubica:
Allow the users to change the settings of parameters within a `PhysicalModel`.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -23,8 +23,8 @@ class ParameterizedModel:
     ----------
     setters : `list` of `tuple`
         A dictionary to information about the setters for the parameters in the form:
-        (name, ParameterSource, setter information). The attributes are stored in the
-        order in which they need to be set.
+        (name, ParameterSource, setter information, required). The attributes are
+        stored in the order in which they need to be set.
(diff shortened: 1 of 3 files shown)


src/tdastro/base_models.py
```diff
@@ -63,39 +62,81 @@ def add_parameter(self, name, value=None, required=False, **kwargs):
         cannot be found.
         Raise a ``ValueError`` if the parameter is required, but set to None.
         """
-        if hasattr(self, name) and getattr(self, name) is not None:
-            raise KeyError(f"Duplicate parameter set: {name}")
+        # Check for parameter has been added and if so, find the index.
+        try:
+            ind = next(ind for ind, entry in enumerate(self.setters) if entry[0] == name)
```
hombit, line 67: Wouldn't dictionary do the `self.setters`' job better? Dictionary guarantees to iterate in the insertion order.

hombit (approved):

src/tdastro/base_models.py
jeremykubica: I didn't know dictionaries were guaranteed to maintain insertion order. Let me merge this PR and send you a followup to change that over.

PR #28, opened 2024-07-12
Change setters list to dict

jeremykubica:
Change the setters list to a dictionary to make the look ups faster. As Kostya noted in #27, dictionaries preserve the order in which the items were inserted.


hombit (approved):
Looks good

PR #29, opened 2024-07-12
Wrap sncosmo models

jeremykubica:
Import the sncosmo model's library and wrap the models in a `PhysicalModel`. This is less efficient  because of the layer of indirection, however it provides flexibility while we port over essential models to be native `PhysicalModel`s. The long term plan will be to port over all needed models so there is not a direct dependency.


mi-dai (commented):
This looks good and would be useful for short term. But I think what we really want to wrap is the Sources in SNCosmo - e.g. https://github.com/sncosmo/sncosmo/blob/903706066adb390d45e33655e466fca402d340e0/sncosmo/models.py#L695
The difference between Model and Source is Source is rest frame SED, while Model includes other effects like redshift.

jeremykubica:
That makes sense. I changed it over to `Source` can you take another look?

mi-dai (approved):
Looks good!

PR #30, opened 2024-07-12
Create a population level model

jeremykubica:
Create a class to store and sample from populations of `PhysicalModel`s.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -2,7 +2,6 @@
 
 import types
 from enum import Enum
-from os import urandom
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


hombit (approved):
It looks good!

I have two general comments:
1. Maybe we should start thinking about switching from `numpy` to `jax` at this point - this code uses `numpy.random` which doesn't exist in `jax`
2. It is not clear if we really want multiple sources simulation. It would also help if you can make a test doing a simple e2e simulation: for example sample a pair of objects and evaluate them for a few times and wavelengths.

mi-dai (commented):
General comment: similar to @hombit's point, I'm not quite sure if we need generate multiple sources altogether. If we do, then the rate model is likely more complicated then a ratio.

hombit (approved):

PR #31, opened 2024-07-12
Create the ability to sample from a background model

jeremykubica:
Creates the concept of a background to a `PhysicalModel` which is just another `PhysicalModel` that adds flux. Illustrate this was a spherical galaxy whose brightness is a Gaussian function of distance from the center. The source is sampled based on the galaxy's parameters and the contributed flux depends on the relative position of the source and galaxy center.

### DIFF
--- a/src/tdastro/sources/galaxy_models.py
+++ b/src/tdastro/sources/galaxy_models.py
@@ -31,7 +31,7 @@ class GaussianGalaxy(PhysicalModel):
         Returns
         -------
         ra : `float`
-            The sampled right ascension.
+            The sampled right ascension in degrees.
         """
         return np.random.normal(loc=self.ra, scale=self.galaxy_radius_std)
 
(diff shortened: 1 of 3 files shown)


src/tdastro/sources/galaxy_models.py
```diff
@@ -0,0 +1,89 @@
...
+        """Draw effect-free observations for this object.
+
+        Parameters
+        ----------
+        times : `numpy.ndarray`
+            A length T array of timestamps.
+        wavelengths : `numpy.ndarray`, optional
+            A length N array of wavelengths.
+        ra : `float`, optional
+            The right ascension of the observations.
```
mi-dai, line 58: "in degree"?

mi-dai (changes requested):
Looks good to me! Need minor change to specify units for ra/dec in docstring to avoid confusions/mistakes

mi-dai:
Another comment: do we want to allow multiple backgrounds? or is it better to have all background calculated in one class?

mi-dai (approved):

PR #33, opened 2024-07-15
Add FunctionNode and basic cosmology

### LINKED ISSUE #11: add cosmology model
add cosmology model to generate observed magnitude based on given absolute magnitude

jeremykubica:
This is a relatively substantial PR that changes how the TDAstro DAG works in a few import ways:
1) `ParameterizedNode`s have a clearer list of what type of setters they accept and cleaner logic to determine which type to use.
2) Create a new `FunctionNode` class to the evaluation DAG that allows users to wrap a function, such as random number generation or cosmology computation. The `FunctionNode` is based off the `ParameterizedNode` so that it keeps sampling in sync when evaluating the function.
3) Add a redshift computation function to both support cosmology and demonstrate how the new function nodes work.


src/tdastro/astro_utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
...
+        astropy.cosmology.units.redshift_distance.
+
+    Returns
+    -------
+    distance : `float`
+        The distance (in pc)
+    """
+    z = redshift * cu.redshift
+    distance = z.to(u.pc, cu.redshift_distance(cosmology, kind=kind))
+    return distance.value
```
hombit, line 28: I believe we need no other type of distance than luminosity distance. I propose to use `distance` parameter name everywhere, while specifying in the docs that it is "luminosity distance"

tests/tdastro/sources/test_physical_models.py
```diff
@@ -1,10 +1,31 @@
...
-    model1 = PhysicalModel(ra=1.0, dec=2.0, distance=3.0)
+    model1 = PhysicalModel(ra=1.0, dec=2.0, distance=3.0, redshift=0.0)
     assert model1.ra == 1.0
     assert model1.dec == 2.0
     assert model1.distance == 3.0
+    assert model1.redshift == 0.0
+
+    # Derive the distance from the redshift using the example from:
+    # https://docs.astropy.org/en/stable/api/astropy.cosmology.units.redshift_distance.html
+    model2 = PhysicalModel(ra=1.0, dec=2.0, redshift=1100.0, cosmology=WMAP9)
```
hombit, line 17: Let's use Planck18 instead

src/tdastro/astro_utils/cosmology.py
jeremykubica: I added the luminosity specifier to the doc strings.

tests/tdastro/sources/test_physical_models.py
jeremykubica: I used it here, but kept the `WMAP9` test in `test_cosmology.py` since I have the answer from the astropy example and can check against that.

src/tdastro/astro_utils/cosmology.py
jeremykubica: As discussed in person, I also changed the astropy function call and removed the `kind` option.

hombit (approved):
Looks good, than k you

OliviaLynn (approved):
Looks good

mi-dai (approved):
Looks good to me!

PR #34, opened 2024-07-15
Improve debugging of TDAstro DAG

jeremykubica:
Three main changes:
1) All the `__str__()` functions are now handled in the base class (`ParameterizedNode`) so that each subclass does not need to repeat them.
2) `ParameterizedNode`s can now have identifiers so we can tell two nodes of the same class apart.
3) Add a helper function to extract all of the attribute settings in the DAG.

### DIFF
--- a/tests/tdastro/test_base_models.py
+++ b/tests/tdastro/test_base_models.py
@@ -156,7 +156,7 @@ def test_parameterized_node_attributes() -> None:
     assert settings["1=PairModel.value2"] == 1.5
     assert settings["1=PairModel.value_sum"] == 2.0
 
-    # Use value1=model.value and value2=1.0
+    # Use value1=model.value and value2=3.0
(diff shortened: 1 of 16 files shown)


tests/tdastro/test_base_models.py
```diff
@@ -141,6 +141,39 @@ def test_parameterized_node() -> None:
...
+    assert settings["value2"] == 1.5
+    assert settings["value_sum"] == 2.0
+
+    settings = model1.get_all_parameter_values(True)
+    assert len(settings) == 3
+    assert settings["1=PairModel.value1"] == 0.5
+    assert settings["1=PairModel.value2"] == 1.5
+    assert settings["1=PairModel.value_sum"] == 2.0
+
+    # Use value1=model.value and value2=1.0
```
OliviaLynn, line 159:
```suggestion
    # Use value1=model.value and value2=3.0
```

OliviaLynn (approved):
Looks good!

PR #37, opened 2024-07-17
Add the ability to set a graph-wide random seed.

jeremykubica:
Add the ability to specify a to set a random seed for all the nodes in a graph. The nodes will each have a specialized seed that is a function of the graph's base seed and their name.

We need to add validity functions to check that all names in a graph are unique.

### DIFF
--- a/src/tdastro/astro_utils/cosmology.py
+++ b/src/tdastro/astro_utils/cosmology.py
@@ -42,16 +42,22 @@ class RedshiftDistFunc(FunctionNode):
         The function or constant providing the redshift value.
     cosmology : `astropy.cosmology`
         The cosmology specification.
+    **kwargs : `dict`, optional
+        Any additional keyword arguments.
     """
 
-    def __init__(self, redshift, cosmology):
+    def __init__(self, redshift, cosmology, **kwargs):
+        # Augment the node identifier string to include the cosmology name.
+        if "node_identifier" in kwargs:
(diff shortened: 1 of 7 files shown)


src/tdastro/base_models.py
```diff
@@ -44,6 +64,27 @@ def __str__(self):
         else:
             return self.__class__.__name__
```
hombit, line 65: I would add a module name, like mentioned in the comment above

hombit (approved):
I do like the random seed approach! I'm a bit worried about `node_identifier` part, but we can make a separate it to a new issue

src/tdastro/base_models.py
jeremykubica: Added

src/tdastro/astro_utils/cosmology.py
```diff
@@ -42,16 +42,22 @@ class RedshiftDistFunc(FunctionNode):
...
+        Any additional keyword arguments.
     """
 
-    def __init__(self, redshift, cosmology):
+    def __init__(self, redshift, cosmology, **kwargs):
+        # Augment the node identifier string to include the cosmology name.
+        if "node_identifier" in kwargs:
+            node_identifier = f"{kwargs['node_identifier']}({cosmology.name})"
+        else:
+            node_identifier = f"{cosmology.name}"
```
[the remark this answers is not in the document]
jeremykubica: I removed this specific override, but didn't do the full meta-programming as we discussed in person.

PR #38, opened 2024-07-17
Extend the __str__() method of FunctionNode

jeremykubica:
We want to __str__() result of a `FunctionNode` to include the function it is calling so we can use multiple `FunctionNode`s in a single graph.


hombit (approved):
Looks great

PR #39, opened 2024-07-18
Add wrappers to random number generators

jeremykubica:
Add random number generator functions from both jax and numpy. These subclass the `FunctionNode` so they can be included directly as nodes in the execution graph. Both node types handle the seeding and maintaining of their respective random number generators.

### DIFF
--- a/src/tdastro/util_nodes/np_random.py
+++ b/src/tdastro/util_nodes/np_random.py
@@ -14,6 +14,20 @@ class NumpyRandomFunc(FunctionNode):
         The name of the random function to use.
     _rng : `numpy.random._generator.Generator`
         This object's random number generator.
+
+    Notes
+    -----
+    Since we need to create a new random number generator for this object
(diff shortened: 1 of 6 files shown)


src/tdastro/util_nodes/np_random.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        super().set_graph_base_seed(graph_base_seed)
+
+        # We create a new random number generator with the new object seed and
+        # link to that object's function.
+        self._rng = np.random.default_rng(self._object_seed)
+        self.func = getattr(self._rng, self.func_name)
+
+    def compute(self, **kwargs):
+        """Execute the wrapped JAX sampling function.
```
hombit, line 62:
```suggestion
        """Execute the wrapped numpy random number generator method.
```

jeremykubica:
As we discussed in person, this implementation doesn't support generating two different independent values within a single node. But I will think about the best way to extend it in order to do that.

PR #42, opened 2024-07-18
Streamline random seed setting

jeremykubica:
Instead of having subclasses override the entire `set_graph_base_seed()` function, break out a smaller, private function that sets the seed and creates and random number generators.


jeremykubica:
Reverting this for now, while I think of a better way to do random numbers,.

PR #43, opened 2024-07-18
Store a list of direct dependencies

jeremykubica:
Store a list of direct dependencies, which are parent nodes in the execution graph. This will allow us to run faster graph traversals that do not require always iterating over every attribute. Regenerate the list of direct dependencies any time an attribute is changes (should be fast since it is not recursive).

Remove the unused `get_dependencies()` function which did the recursive search.


OliviaLynn (approved):
Looks good!

PR #45, opened 2024-07-22
[DRAFT] Proposal for interface simplification

jeremykubica:
An initial proposal to further simplify TDAstro's computation graph interface. Instead of allowing functions (which are opaque to the graph), insist that every attribute is set by: 1) a constant, 2) an attribute of another node, or 3) a function node.


jeremykubica:
Closing because the rest of the interface has changed so much.

PR #47, opened 2024-07-23
Function expansion

jeremykubica:
Several expansions to the TDAstro evaluation graph
1) If a raw function or method is passed as a value, the code will automatically wrap it in a `FunctionNode`.
2) We allow a `FunctionNode` to compute multiple output values.
3) We clear the parameter values when we update the graph information (this will increase (but not completely remove) the likelihood of failing if there is a loop in the execution graph)


hombit (approved):
Looks good

PR #49, opened 2024-07-24
Change how we access another node's attributes

jeremykubica:
Change how both node's parameters are stored in the `ParameterizedNode` objects and how they are access. Instead of storing each parameter as an object attribute (and using `getattr()` and `setattr()`), we now store the parameter values in a dictionary called `parameters`.

**Why?** Being able to access the parameter's directly could lead to confusion during chaining and accidental setting of attributes. For example doing something like `source.ra = 10.0` could cause a problem is the source's RA is supposed to be set from the host. Similarly saying `x1=HostmassX1Func(host.hostmass)` would set `x1` as the current value of the hostmass instead of linking to the object's parameter.

Practically this means that now parameters are referenced via the `parameter` dictionary within a node

**OLD**:
```
return np.full((len(times), len(wavelengths)), self.brightness)
```
becomes.

**NEW**:
```
return np.full((len(times), len(wavelengths)), self.parameters["brightness"])
```

while this is more verbose, it is a lot less accident prone.

A `__getitem__()` function is also provided so the parameters can be accessed from outside the object as if it were a dictionary.

**OLD**:
```
assert source.brightness == 10.0
```
becomes.

**NEW**:
```
assert source["brightness"] == 10.0
```

Finally we create an accessor function with the same name as the parameter so
```
node.add_parameter("x", 10)
```
automatically creates a function `node.x()` which will access the value. This allows us to use a natural short hand when chaining attributes, such as:
```
new_obj = SomeComplexObjectType(value1=old_object.value1, value2=other_object.sum_of_values)
```
where `old_object` has a parameter `value1` and `other_object` has a parameter `sum_of_values`. The code will now do the correct thing and link the objects instead of capturing only those parameter's current values.

### DIFF
--- a/src/tdastro/astro_utils/cosmology.py
+++ b/src/tdastro/astro_utils/cosmology.py
@@ -32,9 +32,6 @@ class RedshiftDistFunc(FunctionNode):
     ----------
     cosmology : `astropy.cosmology`
         The cosmology specification.
-    kind : `str`
-        The distance type for the Equivalency as defined by
-        astropy.cosmology.units.redshift_distance.
 
     Parameters
     ----------
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -62,7 +62,7 @@ class ParameterSource(Enum):
     """
 
     CONSTANT = 1
-    MODEL_ATTRIBUTE = 2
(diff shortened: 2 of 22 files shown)


OliviaLynn (approved):
Looks good!

PR #50, opened 2024-07-25
Fix parameter setting for sncosmo models

jeremykubica:
Fix a problem where the parameter setting to a `SncosmoWrapperModel` were not be propagated to the underlying model. We need to call `self.source.set()` with the correct parameters each time the wrapper model itself is updated. This PR makes that automatic.

### DIFF
--- a/tests/tdastro/sources/test_sncosmo_models.py
+++ b/tests/tdastro/sources/test_sncosmo_models.py
@@ -1,17 +1,54 @@
 import numpy as np
 from tdastro.sources.sncomso_models import SncosmoWrapperModel
+from tdastro.util_nodes.np_random import NumpyRandomFunc
 
 
 def test_sncomso_models_hsiao() -> None:
     """Test that we can create and evalue a 'hsiao' model."""
-    model = SncosmoWrapperModel("hsiao")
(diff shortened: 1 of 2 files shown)


tests/tdastro/sources/test_sncosmo_models.py
```diff
@@ -1,17 +1,54 @@
...
+    hist = [0] * 10
+    source_model = model.source
+    for _ in range(10_000):
+        model.sample_parameters()
+        bin = int(source_model.parameters[0] - 2.0)
+        hist[bin] += 1
+
+    # Each bin should have around 1,000 samples.
+    for i in range(10):
+        assert 800 < hist[i] < 1200
```
mi-dai, line 54: will this always be true?

mi-dai (commented):
Looks good to me! One question about the histogram

tests/tdastro/sources/test_sncosmo_models.py
jeremykubica: That's a good point. Since we are using random numbers, this may occasionally fail with a unlucky draw. I added a constant seed to the test, so it will be reproducible.

mi-dai (approved):

PR #56, opened 2024-07-31
Update the core logic for updating parameters

jeremykubica:
This is a prefactoring PR for some JAX related changes. The goal is to modularize how the parameters are set. The changes should not impact current users and (mostly) be behind the scenes.

What is changing:

1) `ParameterSource` classes is extended to track named information like `source_type` or `fixed`. Previously this was tracked in an array entry and accessed with a magic index number. This should make it easier to see what the code is doing and add new attributes to the setters.

2) `ParameterizedNode` attributes can no longer be set by arbitrary methods of other `ParameterizedNode`s. This was preventing us from easily checking that the evaluation graph was a DAG.

3) The results of a `FunctionNode` are now accessed through their attributes. This prevents problems where the same parameter is fed into multiple nodes and resampled each time (in the `compute()` call). Previously this could lead to inconsistent results.

4) Results of a `FunctionNode` are collected via a function `_save_result_parameters()` to make overloading `compute()` easier.

5) Fixed the seed setting logic in the numpy and Jax function modules to account for the changes above.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -64,7 +64,7 @@ class ParameterSource:
         The type of source as defined by the class variables.
         Default = 0
     value : any
-        The information that actuall sets the parameter. Either a constant
+        The information that actually sets the parameter. Either a constant
         or the attribute name of a dependency node.
     dependency : `ParameterizedNode` or None
         The node on which this parameter is dependent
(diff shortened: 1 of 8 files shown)


OliviaLynn (approved):
Looks good!

PR #63, opened 2024-08-06
[WIP] SN Ia + host galaxy example workflow

mi-dai:
Preliminary work flow and notebook for SN Ia + host galaxy science case.


src/tdastro/sources/snia.py
```diff
@@ -0,0 +1,2 @@
+
```
jeremykubica, line 1: This looks like an empty file that you can remove for now.

src/tdastro/sources/snia_host.py
```diff
@@ -0,0 +1,17 @@
+from tdastro.sources.physical_model import PhysicalModel
+
+class SNIaHost(PhysicalModel):
+    """A static source.
+
+    Attributes
+    ----------
+    radius_std : `float`
```
jeremykubica, line 8: The attributes can be deleted from this doc string since you don't use them in this class.

src/tdastro/astro_utils/snia_utils.py
```diff
@@ -0,0 +1,195 @@
...
+from tdastro.base_models import FunctionNode
+import numpy as np
+from scipy.stats.sampling import NumericalInversePolynomial
+from scipy.stats import norm
+
+class HostmassX1Distr:
+    def __init__(self, hostmass):
+        self.hostmass = hostmass
+
+    def _p(self,x1,hostmass = 9.):
```
jeremykubica, line 11: You will need to run the linter to fix some of the formatting.

If you don't have ruff installed, you can use `pip install ruff`

Then just run:
```
ruff format
ruff check --fix
```
and it will autocorrect most of the formatting.

src/tdastro/astro_utils/snia_utils.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _distmod_from_redshift(redshift,H0=73.,Omega_m=0.3):
+    """Compute distance modulus given redshift and cosmology.
+
+    Parameters
+    ----------
+    redshift : `float`
+        The redshift value.
+    cosmology : `astropy.cosmology`
```
jeremykubica, line 83: You don't need `cosmology` in the doc string since it isn't passed in.

You should add a description for `Omega_m` though.

src/tdastro/astro_utils/snia_utils.py
```diff
@@ -0,0 +1,195 @@
...
+            alpha = alpha,
+            beta = beta,
+            m_abs = m_abs,
+            **kwargs,
+        )
+
+class DistModFromRedshift(FunctionNode):
+    """A wrapper class for the _distmod_from_redshift() function.
+
+    Attributes
```
jeremykubica, line 168: You can remove the Attributes section since there are no attributes in this class.

src/tdastro/astro_utils/snia_utils.py
```diff
@@ -0,0 +1,195 @@
...
+        The function or constant providing the redshift value.
+    H0 : constant
+        The Hubble constant.
+    Omega_m : constant
+        The matter density Omega_m.
+    **kwargs : `dict`, optional
+        Any additional keyword arguments.
+    """
+
+    def __init__(self, redshift,H0=73.,Omega_m=0.3, **kwargs):
```
jeremykubica, line 186: Do you expect to have runs where `H0` and `Omega_m` are generated from some distributions? Just curious.

tests/tdastro/sources/test_snia.py
```diff
@@ -0,0 +1,61 @@
...
+        flux = source.evaluate(phase, wavelengths, resample_parameters=True)
+        p = source.get_all_parameter_values(True)
+
+        res['parameter_values'].append(p)
+        res['flux'].append(flux)
+
+    return res
+
+
+test_snia()
```
jeremykubica, line 61: You don't need this line at the bottom unless you want to run it manually. pytest will automatically pick up any function starting with `test_` during its run.

src/tdastro/astro_utils/snia_utils.py
mi-dai: I think we would always specify the cosmological parameters from a simulation perspective. However if we want to do something like inferring the cosmological parameters using simulations to match the data, we would generate those from some distributions.

mi-dai:
Opened a new PR

PR #68, opened 2024-08-13
Add basic getter for node information

jeremykubica:
Create a simple recursive function to assemble various node information such as: all the node labels, all the node strings, or all the node hash values.


OliviaLynn (approved):
Looks good!

PR #69, opened 2024-08-14
Make the JAX random number generation stateless

jeremykubica:
Remove all the random seed information from the `ParameterizedNode` and instead only maintain it in the numpy and JAX random node classes. This removes a lot of complexity in the base class.

JAX random number generation nodes are stateless and requires the user pass in the current keys each time.

Numpy random number generation nodes can be stateful (with their own stored random number generation) or stateless (with a passed in dictionary of the random number generators).

This PR also adds all the plumbing for passing around the random number generators/keys/state. This adds some complexity to the function signatures, but means that the nodes behave statelessly.

### DIFF
--- a/src/tdastro/sources/physical_model.py
+++ b/src/tdastro/sources/physical_model.py
@@ -93,10 +93,6 @@ class PhysicalModel(ParameterizedNode):
         # Reset the node position to indicate the graph has changed.
         self.node_pos = None
 
-    def set_all_graph_positions(self):
-        """Finalize the graph structure by setting the node positions for the current node,
-        its background, and all effects."""
-
     def set_graph_positions(self, seen_nodes=None):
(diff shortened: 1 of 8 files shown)


src/tdastro/sources/physical_model.py
```diff
@@ -90,6 +90,32 @@ def add_effect(self, effect, allow_dups=True, **kwargs):
 
         self.effects.append(effect)
 
+        # Reset the node position to indicate the graph has changed.
+        self.node_pos = None
+
+    def set_all_graph_positions(self):
```
OliviaLynn, line 96: Did this get iterated on and stranded?

jeremykubica: Yes. That was left there by accident.

OliviaLynn (approved):

Issue #61, opened 2024-08-05
Create single "Passband" class


OliviaLynn:
I would vote for implementing a class for a single band, `Passband`, instead. Collection of passbands could also be useful, but I don see how we are going to use it practically.

1) For each observation we need a single passband only, not all of them. It sounds reasonable to me to pass passband object per observation. We also could pass band name, but I see some potential issues with that in the future,
2) we will need a more flexible interface for passbands which will be able to change transmission with airmass for each individual observation (I think it is what Andy C asked us during LINCC UP). That means that passband name ID will not be enough, because transmission will be parameterized by both filter and airmass value.
3) It would be harder co combine surveys if we have a single entry point for all passbands

_Originally posted by @hombit in https://github.com/lincc-frameworks/tdastro/pull/52#discussion_r1698380338_

OliviaLynn:
![Screenshot 2024-08-23 at 1 27 36 PM](https://github.com/user-attachments/assets/146919fd-4fb7-4883-b46e-c4742ae4ea7b)

PR #70, opened 2024-08-20
Passband and PassbandGroup classes

OliviaLynn:
For #61 

### Summary
- Provides individual `Passband` class, which can be bundled up in a `PassbandGroup` class 
- `Passband`:
  - holds metadata relevant to that passband (eg, label="u", survey="LSST")
  - holds data such as the passband's transmission table (and handles the loading/downloading required for this)
  - holds all of the equations referenced from the LSST Flux Units paper
- `PassbandGroup`:
  - holds all of the `Passband`s
  - can be initialized with presets such as "LSST", which automatically constructs the appropriate `Passband` objects
  - provides a `fluxes_to_bandfluxes` method to easily generate all bandfluxes at once for a given flux density matrix
  
### Disclaimers
- This PR will be followed up with interpolation and exterpolation utilities. 
  - Note the checks in the beginning of `Passband`'s `fluxes_to_bandflux`--this will soon be made much more robust than it is now.
- The downloading functionality is intentionally minimal. 
  - Current download method will be replaced when implementing a Pooch-based data manager; so, only a very basic example of downloading LSST bands from the SVO website was provided, mainly as a matter of convenience during development.


src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
-                return False
+    Attributes
+    ----------
+    label : str
+        The label of the passband.
+     survey : str
+        The survey to which the passband belongs.
+    full_name : str
+        The full name of the passband.
+    table_path : str
```
jeremykubica, line 101: I'm not sure `table_path` and `table_url` need to be attributes. You could just pass them along to the respective functions during `__init__`. It's not a problem to keep things as-is for now though.

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
 
-            logging.info(f"Downloaded: {band_name}-band transmission table.")
-            return True
-        except HTTPError as e:
-            logging.error(f"HTTP error occurred when downloading transmission table for {band_name}: {e}")
-            return False
-        except URLError as e:
-            logging.error(f"URL error occurred when downloading transmission table for {band_name}: {e}")
-            return False
+    def __init__(self, survey, label, table_path=None, table_url=""):
```
jeremykubica, line 111: Would it be useful to have the option of passing in the transmission table as a numpy array manually? I can imagine cases where that would be useful for testing, but it might not be needed in normal runs of the code.

jeremykubica (approved):
The code looks a lot cleaner broken up into two classes. I have a few optional suggestions.

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
         -------
         bool
-            True if the transmission table is successfully loaded, False otherwise.
+            True if the download was successful, False otherwise.
         """
+        if self.table_url == "":
+            if self.survey == "LSST":
+                # TODO: This area will be changed with incoming Pooch data manager PR :)
+                # Consider: https://github.com/lsst/throughputs/blob/main/baseline/filter_g.dat
+                # Or check with Mi's link (unless that was above)
```
mi-dai, line 150: I think we want to use this one: https://github.com/lsst/throughputs/blob/main/baseline/total_g.dat

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
+                f"Extrapolation needed: flux density interval [{wavelengths_angstrom[0]}, "
+                f"{wavelengths_angstrom[-1]}] is outside the transmission table interval "
+                f"[{self.normalized_transmission_table[0, 0]}, {self.normalized_transmission_table[-1, 0]}]."
+            )
 
-            # Find the indices of the wavelengths in the band in the union array
-            band_indices = np.searchsorted(all_wavelengths, wavelengths_in_band)
+        # Get only the flux densities that are within the passband's wavelength range
+        passband_wavelengths = self.normalized_transmission_table[:, 0]
+        passband_wavelengths_indices = np.searchsorted(wavelengths_angstrom, passband_wavelengths)
```
mi-dai, line 239: I'm curious if it'll be better/faster to search for the min and max of the passband_wavelengths instead of all overlapping indices (since the wavelengths are supposed to be sorted).

mi-dai: On a related note, should we check if the wavelengths are sorted when we load them? (It's likely good to check, but maybe that's the user's responsibility)

mi-dai (approved):
Looks good to me! I thought it would still be useful to have something like `model.get_bandfluxes(passbands, times)` where we wrap these under PhysicalModel - 
```
      def get_bandfluxes(self,...):
             flux = self.evaluate(...)
             bandfluxes = passbands.fluxes_to_bandfluxes(...)
```
maybe we can hold for now.

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
-        if not os.path.exists(data_dir):
-            os.makedirs(data_dir)
-        self.data_path = data_dir
+        self.passbands = {}
+
+        if preset is not None:
+            self._load_preset(preset)
+
+        if passbands is not None:
+            self.passbands = {}
```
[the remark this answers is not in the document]
OliviaLynn: Good point--I must have overlooked the `self.passbands = {}` still on 38 when moving it to 32. 

I'm inclined to take L38 out, to allow someone to load a preset plus any additional bands they might want to compare a survey with, but I think @mi-dai would know best if this is a use case someone might want.

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
+    """A passband contains information about its transmission curve and calculates its normalization.
 
-            if os.path.getsize(output_file) == 0:
-                logging.error(f"Transmission table for {band_name} is empty.")
-                return False
+    Attributes
+    ----------
+    label : str
+        The label of the passband.
+     survey : str
```
OliviaLynn, line 97:
```suggestion
    survey : str
```

src/tdastro/astro_utils/passbands.py
OliviaLynn: I was on the fence about that--ultimately I chose to make the unit tests rather than the code be slightly more complicated, but would be happy to switch if that's preferred.

src/tdastro/astro_utils/passbands.py
```diff
@@ -2,257 +2,246 @@
...
+        The survey to which the passband belongs.
```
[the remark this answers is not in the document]
OliviaLynn: I think it is better if it is always included--I understand a u-band from LSST, for instance, is different from a u-band from HSC. 
This has made me realize I've been using "label" (eg, "u") as my dict key, so I've changed it over to the more appropriate/actually unique "full_name" (eg, "LSST_u").

src/tdastro/astro_utils/passbands.py
OliviaLynn: - Good point--I'll use min and max for extrapolation check in the following PR. 
- I agree with you on sorted wavelengths--I recently put that in the function where I check the grid steps/if we need to interpolate at all; it's nice not to make users stay on top of that if we can do it too

src/tdastro/astro_utils/passbands.py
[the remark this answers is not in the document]
mi-dai: I think we'd better not have passbands and preset all set to values to avoid confusion.

PR #71, opened 2024-08-20
Add **

OliviaLynn:
Was masking the arguments needed for resampling in redshift effects (and presumably others)


OliviaLynn:
Wait actually @mi-dai let me cross check this with the changes we made locally on your machine

PR #72, opened 2024-08-22
SN Ia + host galaxy example workflow

mi-dai:
Simulating SN Ia light curves using the SALT model in sncosmo. Added toy relation between x1 and hostmass. Simple opsim applied for cadence (no filter info).


mi-dai:
Closing this one because I have too many unrelated histories

PR #74, opened 2024-08-23
Add a t0 parameter to sncosmo models

jeremykubica:
Add a t0 parameter to sncosmo models. Like all parameter this can be sampled if a function is passed in.

Also expands the comments to indicate which times are in rest frame vs observer frame.

### DIFF
--- a/tests/tdastro/sources/test_sncosmo_models.py
+++ b/tests/tdastro/sources/test_sncosmo_models.py
@@ -5,35 +5,46 @@ from tdastro.util_nodes.np_random import NumpyRandomFunc
 
 def test_sncomso_models_hsiao() -> None:
     """Test that we can create and evalue a 'hsiao' model."""
-    model = SncosmoWrapperModel("hsiao", amplitude=1.0e-10)
+    model = SncosmoWrapperModel("hsiao", amplitude=2.0e10)
     state = model.sample_parameters()
-    assert model.get_param(state, "amplitude") == 1.0e-10
+    assert model.get_param(state, "amplitude") == 2.0e10
     assert model.get_param(state, "t0") == 0.0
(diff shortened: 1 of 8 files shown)


tests/tdastro/sources/test_sncosmo_models.py
```diff
@@ -18,6 +19,18 @@ def test_sncomso_models_hsiao() -> None:
     assert np.allclose(fluxes, [4.31210900e-20, 7.46619962e-20, 1.42182787e-19])
```
mi-dai, line 19: I think I missed this one last time I reviewed it. The example on the sncosmo site has set the model t0 = 55000 but we have t0=0 (so we will get zero fluxes) and they shouldn't match. The reason the test passed is because the flux values are very small. The default threshold for np.allclose is 1e-8

tests/tdastro/sources/test_sncosmo_models.py
```diff
@@ -18,6 +19,18 @@ def test_sncomso_models_hsiao() -> None:
...
 
 
+def test_sncomso_models_hsiao_t0() -> None:
+    """Test that we can create and evalue a 'hsiao' model with a nonzero t0."""
+    model = SncosmoWrapperModel("hsiao", amplitude=1.0e-10, t0=10.0)
```
mi-dai, line 24: same as above, we should have `t0=55000` here. and `[54990]` in L30 (but the fluxes still won't match because redshift=0.5)

tests/tdastro/sources/test_sncosmo_models.py
mi-dai: Also the example is setting z=0.5, so we would not get a matched flux without the redshift effect.

mi-dai (changes requested):

mi-dai (approved):

PR #75, opened 2024-08-23
Optimize the OpSim Search

jeremykubica:
Wrap the opsim functions in a class OpSim and prebuild an internal kd-tree to do fast lookups of of whether a point is included within the survey. Most of the actual code is unchanged, but shows up as a change because it is now part of a class.

### DIFF
--- a/src/tdastro/astro_utils/opsim.py
+++ b/src/tdastro/astro_utils/opsim.py
@@ -3,146 +3,196 @@ from pathlib import Path
 
 import numpy as np
 import pandas as pd
-from astropy.coordinates import angular_separation
-from astropy.table import Table
+from scipy.spatial import KDTree
+
+_rubin_opsim_colnames = {
+    "time": "observationStartMJD",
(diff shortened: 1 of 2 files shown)


src/tdastro/astro_utils/opsim.py
```diff
@@ -3,146 +3,196 @@
...
+        """Return the indices of the opsim pointings that fall within the field
+        of view of the query point(s).
+
+        Parameters
+        ----------
+        query_ra : `float` or `numpy.ndarray`
+            The query right ascension (in degrees).
+        query_dec : `float` or `numpy.ndarray`
+            The query declination (in degrees).
+        fov : `float`
```
hombit, line 148: Usually we think about fov to be a square root of area or diameter or edge length  we shouldn't define it as radius, I'd prefer to use diameter here. Also we may use LSST value as a default, diameter is 3.5 degrees

hombit (approved):
Looks great. I suggest some changes of FoV definition.

PR #79, opened 2024-08-27
Enable basic benchmarks for TDAstro

jeremykubica:
Uses the Python Project Template to turn on benchmarking for TDAstro and adds a single benchmark.


camposandro (approved):
Looks good! I would just recommend deleting the `gh-pages` branch before merging the pull request because that branch is going to be created and managed by the workflows that deploy the asv dashboard.

PR #81, opened 2024-08-28
Add a test OpSim file

jeremykubica:
Creates a very small and simple OpSim file that can be use by tests.

Includes the script for generating the OpSim file in tests/tdastro/util so we can easily regenerate it in the future.


tests/tdastro/utils/make_fake_opsim.py
```diff
@@ -0,0 +1,84 @@
...
+            for dec in dec_vals:
+                opsim_data["observationStartMJD"].append(t)
+                opsim_data["fieldRA"].append(ra)
+                opsim_data["fieldDec"].append(dec)
+                t += time_step
+
+    # Add other fields
+    num_obs = len(opsim_data["observationStartMJD"])
+    opsim_data["zp_nJy"] = np.ones(num_obs)
+    opsim_data["filter"] = np.full((num_obs), "r")
```
mi-dai, line 44: Can we add another filter? 
```suggestion
    opsim_data["filter"] = np.full((num_obs), "r")
    opsim_data["filter"][0::2] = "g"
```

mi-dai (approved):

PR #82, opened 2024-08-28
Extend OpSim to filter on values in given columns

jeremykubica:
Add functionality for the `OpSim` object to filter a list of indices on a given value. Use this new functionality to add filter matching for range_search.


hombit:
I need some context on why do we need these changes

PR #89, opened 2024-08-29
Adapt HostmassX1Func to use random node infrastructure

jeremykubica:
Create a new `NumericalInversePolynomialFunc` node that wraps scipy's `NumericalInversePolynomial`. This wrapping allows us to reuse all the of the infrastructure for getting and setting seeds.

Change `HostmassX1Func` to use this `NumericalInversePolynomialFunc`. Add a bunch of tests to confirm that it is generating both new samples and underlying hostmasses.


hombit (approved):
It looks fine. I think we potentially could do it faster (and maybe less precise) with `qp` or our own implementation, but I think it is fine to go with this for now.

PR #92, opened 2024-08-30
Cleanup and simplify internal logic

jeremykubica:
The goal of this PR is to cleanup and simplify the internal logic for sampling so it is easier to understand and debug. To this end, it:
- Adds a `__str__()` function to `GraphState` that produces a human readable debugging string. 
- Replaces the nodes' hash value with the nodes' (readable) string in `GraphState`. This will add slight overhead in indexing (to hash the string), but greatly increase testability and debug-ability.
- Switch the PyTree to use the same dictionary structure as `GraphState` (a double index of [node_name][var_name]). This allows:
    - Removing the now unused `full_name` attribute from `ParameterSource`
    - Removing a lot of specialized logic that was needed to handle `given_args`. Instead the `given_args` are merged directly into the current `GraphState` as fixed parameters.

### DIFF
--- a/src/tdastro/base_models.py
+++ b/src/tdastro/base_models.py
@@ -463,7 +463,7 @@ class ParameterizedNode:
         getter.__name__ = name
         setattr(self, name, getter)
 
-    def compute(self, graph_state, given_args=None, rng_info=None, **kwargs):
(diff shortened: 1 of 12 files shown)


OliviaLynn (approved):
Looks good! Just the one error message

PR #97, opened 2024-09-03
LSST point-source noise model

hombit:
This adds
- `astro_utils.noise_model` module with a simple function `poisson_flux_std` for noise computation
- `astro_utils.opsim.OpSim.noise_point_source` method which produces `flux_err` from band flux and opsim data.

This PR doesn't implement anything for applying noise to flux, e.g. doesn't add randomly sampled noise to the original "perfect" band flux.

I test it for OpSim's `fiveSigmaDepth` and got a difference of up to ~15%, which was also the case with zeropoint test. We should investigate the root of this difference in the future.

Here, I implement a “textbook” noise computation, but perhaps we are actually more interested in “noise simulation” instead. By “noise simulation,” I mean a pipeline that takes the perfect band flux and simulates the electron count using a Poisson distribution. Then, we convert the counts back to flux units and output the band flux along with its error estimation. In the case of the code in the PR, this error estimation currently equals the Poisson standard deviation, but other noise factors (e.g., zero-point error) may change the flux error distribution. Let’s discuss this further later!​

## Change Description

- [ ] My PR includes a link to the issue that I am addressing

## Code Quality
- [ ] I have read the Contribution Guide
- [ ] My code follows the code style of this project
- [ ] My code builds (or compiles) cleanly without any errors or warnings
- [ ] My code contains relevant comments and necessary documentation

### Bug Fix Checklist
- [ ] My fix includes a new test that breaks as a result of the bug (if possible)
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### New Feature Checklist
- [ ] I have added or updated the docstrings associated with my feature using the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)
- [ ] I have updated the tutorial to highlight my new feature (if appropriate)
- [ ] I have added unit/End-to-End (E2E) test cases to cover my new feature
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### Documentation Change Checklist
- [ ] Any updated docstrings use the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)

### Build/CI Change Checklist
- [ ] If required or optional dependencies have changed (including version numbers), I have updated the README to reflect this
- [ ] If this is a new CI setup, I have added the associated badge to the README

### Other Change Checklist
- [ ] Any new or updated docstrings use the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the tutorial to highlight my new feature (if appropriate)
- [ ] I have added unit/End-to-End (E2E) test cases to cover any changes
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### DIFF
--- a/src/tdastro/astro_utils/opsim.py
+++ b/src/tdastro/astro_utils/opsim.py
@@ -8,12 +8,13 @@ import pandas as pd
 from scipy.spatial import KDTree
 
 from tdastro.astro_utils.mag_flux import mag2flux
-from tdastro.astro_utils.noise_model import GAUSS_EFF_AREA2FWHM_SQ, poisson_flux_std
+from tdastro.astro_utils.noise_model import poisson_flux_std
 from tdastro.astro_utils.zeropoint import (
     _lsstcam_extinction_coeff,
     _lsstcam_zeropoint_per_sec_zenith,
     flux_electron_zeropoint,
 )
+from tdastro.consts import GAUSS_EFF_AREA2FWHM_SQ
 
 _rubin_opsim_colnames = {
     "time": "observationStartMJD",
--- a/src/tdastro/astro_utils/opsim.py
+++ b/src/tdastro/astro_utils/opsim.py
@@ -309,7 +309,7 @@ class OpSim:  # noqa: D101
         flux_err : array_like of float
             Simulated flux noise in nJy.
         """
-        observations = self.table[index]
+        observations = self.table.iloc[index]
 
         # By the effective FWHM definition, see
         # https://smtn-002.lsst.io/v/OPSIM-1171/index.html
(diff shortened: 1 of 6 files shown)


src/tdastro/astro_utils/opsim.py
```diff
@@ -291,3 +293,39 @@ def get_observations(self, query_ra, query_dec, radius, cols=None):
                 raise KeyError(f"Unrecognized column name {table_col}")
             results[col] = self.table[table_col][neighbors].to_numpy()
         return results
+
+    def flux_err_point_source(self, flux, index):
+        """Simulate photon noise for a point source observation
```
jeremykubica, line 298: I was a little confused by the comment when I first read this, because I thought it would be sampling from the noise distribution. Maybe we should say something like "Compute the distribution parameters of photon noise..."

jeremykubica (approved):
Two small stylistic questions.

PR #99, opened 2024-09-04
Update the small opsim db

### LINKED ISSUE #98: Update test OpSim file to match columns to the original file
Could we please update `opsim_small.db` test file to match all the columns from the original opsim file. It would require both adding all the missed columns and removing `zp_nJy` which we generate with the `OpSim` class.

This changes are needed to write tests for PR #97

jeremykubica:
Closes #98 

Update the opsim test data to match the columns in the Rubin data. In order to generate the new data without information like AirMass, we need to make adding the zero points optional. This PR changes zero point addition from happening automatically in the constructor to being a manual call of `add_zero_points()`.


hombit:
I'll make a PR for this PR after a (conflicting) #97 merged

hombit (approved):

PR #107, opened 2024-09-10
SN Ia rates

mi-dai:
Add functions to calculate number of SNe in redshift bins based on their volumetric rates.

### DIFF
--- a/src/tdastro/astro_utils/snia_utils.py
+++ b/src/tdastro/astro_utils/snia_utils.py
@@ -70,7 +70,7 @@ def num_snia_per_redshift_bin(zmin=0.001, zmax=10, znbins=20, solid_angle=None, 
 
     cosmo = FlatLambdaCDM(H0=H0, Om0=Omega_m)
 
-    zarr = np.linspace(zmin, zmax, znbins)
+    zarr = np.linspace(zmin, zmax, znbins + 1)
 
     int_arr = np.linspace(zarr[:-1], zarr[1:], 50, axis=1)
     z_mean = np.mean(int_arr, axis=1)
--- a/tests/tdastro/sources/test_snia_rates.py
+++ b/tests/tdastro/sources/test_snia_rates.py
@@ -26,7 +26,7 @@ def test_num_snia_per_redshift_bin():
     # zmean = 0.073
     zmin = 0.07
     zmax = 0.08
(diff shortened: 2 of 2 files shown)


src/tdastro/astro_utils/snia_utils.py
```diff
@@ -7,6 +8,84 @@
...
+    """
+
+    r0 = 2.27e-5
+    alpha = 1.7
+    rate_vol = r0 * np.power(1.0 + redshift, alpha)
+
+    return rate_vol
+
+
+def num_snia_per_redshift_bin(zmin=0.001, zmax=10, znbins=20, solid_angle=None, H0=73.0, Omega_m=0.3):
```
hombit, line 36: I don't really get why do we need bin function. For me it looks like `snia_volumetric_rates(redshift) * dV(redshift)` should be enough to build a redshift distribution to sample from...

mi-dai: I haven't thought carefully on how this should be used. I think there are two things that are needed: 1. a redshift distribution (which is likely a function of all host properties) 2. The total number of SNe (which also depends on length, solid angle)

hombit (approved):

OliviaLynn (approved):
Looks good! One comment, I would expect the znbins parameter to correspond to the number of bins we generate. 

It seems that in using it to generate the boundaries for the bins, we generate 1 fewer bin. With the unit tests basically acknowledging this, I wonder if this is a standard convention in astronomy--but in case it isn't, I might add a note in the docstrings, or change L73 from `zarr = np.linspace(zmin, zmax, znbins)` to `zarr = np.linspace(zmin, zmax, znbins+1)`

mi-dai:
> Looks good! One comment, I would expect the znbins parameter to correspond to the number of bins we generate.
> 
> It seems that in using it to generate the boundaries for the bins, we generate 1 fewer bin. With the unit tests basically acknowledging this, I wonder if this is a standard convention in astronomy--but in case it isn't, I might add a note in the docstrings, or change L73 from `zarr = np.linspace(zmin, zmax, znbins)` to `zarr = np.linspace(zmin, zmax, znbins+1)`

Good call! I realized they are different when writing the unit test but didn't think much more of it. I'll get it changed.

PR #109, opened 2024-09-13
Increate the flexibility of the OpSim Table.

jeremykubica:
Make `OpSim` easier to use for testing:

- Provide a more meaningful error if zero points are not provided.
- Add the helper method `has_columns()` that wraps the column name map.
- Add a function to add columns with column name map support.
- Add the function `opsim_add_random_data()` to create a new column filled with random data (for testing).

### DIFF
--- a/src/tdastro/astro_utils/opsim.py
+++ b/src/tdastro/astro_utils/opsim.py
@@ -357,3 +357,23 @@ class OpSim:  # noqa: D101
             readout_noise=self.read_noise,
             dark_current=self.dark_current,
         )
+
+
+def opsim_add_random_data(opsim_data, colname, min_val=0.0, max_val=1.0):
+    """Add a column composed of random uniform data. Used for testing.
+
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


hombit (commented):
It's look good, but I'm still opposite the idea of making zero-point calculations optional. 

The technical reason is that it allows `OpSim` to be in an invalid state, when we cannot really run `flux_err_point_source()`.

The pipeline reason is that we do need all these columns to be there for any simulation, because there is no point of doing any simulation without error estimation. I'd vote for fine-tuning `OpSim` class for LSST purposes, at least for now, while trying to separate any potential common functionality to separate modules (like with `noise_model`, `passband`, etc)

jeremykubica:
I reverted the optionality of the zero point computation for now, but we should discuss at Tuesday's meeting. I think we will have a variety of things we want to compute from `OpSim` and only need a subset for specific tasks.

hombit (approved):
Thank you!

Issue #76, opened 2024-08-23
Move redshift to Physical Model


OliviaLynn:
> Instead of applying redshift as an effect, I think we should make it part of PhysicalModel. If redshift != 0, PhysicalModel.evaluate always calls the pre-effect function to transform the frame and then always calls the post effect afterward. This way the user does not need to manually add redshift effects.

PR #111, opened 2024-09-16
Passband/PassbandGroup new interface

### LINKED ISSUE #64: Passband interpolation
I think we might want to define a integration step, and interpolate both the model SED and the transmission curve (and potentially allow the step-size to be changed)

_Originally posted by @mi-dai in https://github.com/lincc-frameworks/tdastro/pull/52#discussion_r1700821156_

OliviaLynn:
Passband and PassbandGroup now follow a new interface:
```
source = Source(...)
passband = Passband(delta_wave: float | None = 5, trim_percentile: float | None = 0.1)
spectral_flux = source.evaluate(passband.waves, t)
band_flux = passband.flux_to_bandflux(spectral_flux)
```
### Summary
After some iteration, we decided we could avoid the need for any flux interpolation/extrapolation if we:
1. Initialize our Passband (or PassbandGroup) object early on
2. Get the wavelengths used in its transmission table(s)
3. Feed those wavelengths to our model's `evaluate` method
4. Generate a spectral flux/flux density matrix that is already on the same grid (and shares the same interval) as our target passband

**Note** that in step 1, we can also specify a target grid step we would like to be using (additionally, we can fine-tune trimming the tails of our transmission tables via the `trim_quantile` parameter).

Both of these parameters may be reset at any time via `process_transmission_tables`, which will re-interpolate the transmission table(s) as needed.

### Note:
- I have updated the snia unit test to conform to this new interface, and it runs successfully when ```redshift=NumpyRandomFunc("uniform", low=0.01, high=0.02),``` is replaced with ```redshift=0```. 
I do not believe the failure here to be related to passbands. 
- See second-to-last commit (or, second-to-last-before-PR-comment-commits) https://github.com/lincc-frameworks/tdastro/pull/111/commits/805e8a7f68e6dee6b5043a9be3a5ab60197d48ff for unit tests passing with this change made (however, I have since returned redshift back to NumpyRandomFunc).
- Considering the best course of action to be addressing this as part of the scope of https://github.com/lincc-frameworks/tdastro/issues/76, but open to suggestion

### Issue
Will close #64


tests/tdastro/astro_utils/test_passbands.py
```diff
@@ -116,51 +102,161 @@ def mock_urlretrieve(url, filename, *args, **kwargs):  # Urlretrieve saves conte
...
+    )
+    c_band = create_test_passband(tmp_path, transmission_table, filter_name="c")
+
+    # Trim the transmission table by 5% on each side
+    table = c_band._trim_transmission_by_percentile(c_band._loaded_table, trim_percentile=0.05)
+    assert len(table) < len(c_band._loaded_table)
+
+    original_area = np.trapz(c_band._loaded_table[:, 1], x=c_band._loaded_table[:, 0])
+    trimmed_area = np.trapz(table[:, 1], x=table[:, 0])
+    assert np.isclose(trimmed_area, original_area * 0.9, rtol=0.01)
```
hombit, line 230: Let's maybe check that trimmed area is `>= original_area * 0.9` instead?

hombit (commented):
Great, thank you so much! Everything looks great, the only large thing is actually my fault and asks to rename `trim_percentile` everywhere.

src/tdastro/astro_utils/passbands.py
```diff
@@ -102,134 +168,149 @@ class Passband:
...
+            The filter_name of the passband: eg, "u".
+        delta_wave : float or None, optional
+            The grid step of the wave grid. Default is 5.0.
+        trim_percentile : float or None, optional
+            The percentile to trim the transmission table by. For example, if trim_percentile is 0.1, the
+            transmission table will be trimmed to include only the central 80% of the area under the
+            transmission curve.
+        waves : np.ndarray
+            The wavelengths of the transmission table. To be used when evaluating models to generate fluxes
+            that will be passed to fluxes_to_bandflux.
```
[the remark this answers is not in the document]
OliviaLynn: Good catch! It should just be in Attributes, not also in Parameters

src/tdastro/astro_utils/passbands.py
```diff
@@ -242,31 +323,155 @@ def _download_transmission_table(self, table_path: Path, table_url: str) -> bool
...
+        transmissions = table[:, 1]
+
+        # Calculate the cumulative sum of the transmission values (area under the curve)
+        cumulative_area = scipy.integrate.cumulative_trapezoid(transmissions, x=wavelengths)
+
+        # Normalize cumulative area to range from 0 to 1
+        cumulative_area /= cumulative_area[-1]
+
+        # Find indices where the cumulative area exceeds the trim percentiles
+        lower_bound = np.searchsorted(cumulative_area, trim_percentile)
```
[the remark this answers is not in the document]
OliviaLynn: That makes sense! I'm implementing this a max() because I think that's what you meant for this

src/tdastro/astro_utils/passbands.py
```diff
@@ -18,36 +17,55 @@ class PassbandGroup:
...
+        **kwargs
+            Additional keyword arguments to pass to the Passband constructor.
         """
         self.passbands = {}
 
         if preset is not None:
-            self._load_preset(preset)
+            self._load_preset(preset, **kwargs)
+
+        if passband_parameters is not None:
```
jeremykubica, line 59: Do you want to raise an error if this is None? Or does the class use reasonable defaults in that case?

src/tdastro/astro_utils/passbands.py
```diff
@@ -56,44 +74,92 @@ def __str__(self) -> str:
...
+            )
+
         bandfluxes = {}
-        for label, passband in self.passbands.items():
-            bandfluxes[label] = passband.fluxes_to_bandflux(flux_density_matrix, wavelengths_angstrom)
+        for full_name, passband in self.passbands.items():
+            # We only want the fluxes that are in the passband's wavelength range
+            # So, find the indices in the group's wave grid that are in the passband's wave grid
+            lower, upper = passband.waves[0], passband.waves[-1]
+            lower_index, upper_index = np.searchsorted(self.waves, [lower, upper])
```
jeremykubica, line 151: This looks like it isn't unique per function call. If so, it would be more efficient to do these operations once when setting up the data structure. That would mean keeping a separate dictionary of full_name to (lower_index, upper_index), so you'd need to look at how much time it saves for the added complexity.

src/tdastro/astro_utils/passbands.py
```diff
@@ -56,44 +74,92 @@ def __str__(self) -> str:
...
+            # Check that this is the right grid after all (check will fail if the grid is not regular, or
+            # passbands are overlapping)
+            if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):
```
jeremykubica, line 154: Again, I think this could be precomputed during data creation and stored in a dictionary mapping full_name to a boolean indicating that the arrays are equal.

src/tdastro/astro_utils/passbands.py
```diff
@@ -102,134 +168,149 @@ class Passband:
...
     def __init__(
         self,
         survey: str,
-        label: str,
+        filter_name: str,
+        delta_wave: Optional[float] = 5.0,
+        trim_quantile: Optional[float] = 1e-3,
         table_path: Optional[str] = None,
         table_url: Optional[str] = None,
         units: Optional[Literal["nm", "A"]] = "A",
```
jeremykubica, line 200: Should we just use astropy units here so we can use all their conversions?

src/tdastro/astro_utils/passbands.py
```diff
@@ -102,134 +168,149 @@ class Passband:
...
+        delta_wave : float or None, optional
+            The grid step of the wave grid, in angstroms.
+            It is typically used to downsample transmission using linear interpolation.
+            Default is 5 angstroms. If `None` the original grid is used.
+        trim_quantile : float or None, optional
+            The quantile to trim the transmission table by. For example, if trim_quantile is 1e-3, the
+            transmission table will be trimmed to include only the central 99.8% of the area under the
+            transmission curve.
         table_path : str, optional
-            Path to the table defining the passband on the filesystem. Will take precedence over table_url
```
jeremykubica, line 138: I think it is fine to keep this as-is, but an alternative approach that might be more modular approach might be to have the constructor take a table and then to add two class methods:

```
@classmethod
def from_file(cls, table_path: str, ...):
   load the file
   ...
   return Passband(table, ...)

@classmethod
def from_url(cls, table_url: str, force_download: bool, ...):
    do the download
    ...
    return Passband(table, ...)
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -242,31 +323,154 @@ def _download_transmission_table(self, table_path: Path, table_url: str) -> bool
...
 
-    def _normalize_transmission_table(self) -> np.ndarray:
+    def process_transmission_table(
+        self, delta_wave: Optional[float] = 5.0, trim_quantile: Optional[float] = 1e-3
+    ):
+        """Process the transmission table.
+
+        Parameters
+        ----------
+        delta_wave : Optional[float] = None
```
jeremykubica, line 333: I don't think the `= None` should be here given that is has a default of 5.0.

src/tdastro/astro_utils/passbands.py
```diff
@@ -242,31 +323,154 @@ def _download_transmission_table(self, table_path: Path, table_url: str) -> bool
...
+            The grid step of the wave grid. Default is 5.0 Angstroms.
+        trim_quantile : Optional[float] = None
```
jeremykubica, line 335: I don't think the `= None` should be here given that is has a default of 1e-3.

src/tdastro/astro_utils/passbands.py
```diff
@@ -242,31 +323,154 @@ def _download_transmission_table(self, table_path: Path, table_url: str) -> bool
...
+        Parameters
+        ----------
+        transmission_table : np.ndarray
+            A 2D array of wavelengths and transmissions.
 
         Returns
         -------
         np.ndarray
             A 2D array of wavelengths and normalized transmissions.
+        Raises
```
jeremykubica, line 440: Add a space before this line.

tests/tdastro/astro_utils/test_passbands.py
```diff
@@ -116,51 +102,164 @@ def mock_urlretrieve(url, filename, *args, **kwargs):  # Urlretrieve saves conte
...
+    # Now call without mocking, to check waves set correctly (other values checked in method-specific tests)
+    delta_wave, trim_quantile = 5.0, None
+    a_band.process_transmission_table(delta_wave, trim_quantile)
+    np.testing.assert_allclose(a_band.waves, np.arange(100, 301, delta_wave))
+
+
+def test_interpolate_transmission_table(tmp_path):
+    """Test the _interpolate_transmission_table method of the Passband class."""
+    transmission_table = "100 0.5\n200 0.75\n300 0.25\n"
+    a_band = create_test_passband(tmp_path, transmission_table)
```
jeremykubica, line 143: One advantage of allowing the constructor to take in a table (see suggestion in passbands.py) would be that you don't have to write out to files in each of these tests, which would speed things up.

src/tdastro/astro_utils/passbands.py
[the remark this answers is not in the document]
hombit: Right, it should be `max()`, thank you!

tests/tdastro/sources/test_snia.py
```diff
@@ -129,11 +129,25 @@ def test_snia_end2end(
...
+                "survey": "LSST",
+                "filter_name": "u",
+                "table_url": f"{url_base}baseline/total_u.dat",
+                "units": "nm",
+                "force_download": True,
+            },
+            {
+                "survey": "LSST",
+                "filter_name": "r",
+                "table_url": f"{url_base}baseline/total_r.dat",
```
jeremykubica, line 143: How large are these files? Could we include them (or a suitable fake) in the git repository? Downloading files from a URL during a test increases the change the test will randomly fail. So we should try to avoid that as much as possible.

jeremykubica (approved):

hombit (approved):
Thank you!

src/tdastro/astro_utils/passbands.py
OliviaLynn: Quoted in #112

src/tdastro/astro_utils/passbands.py
OliviaLynn: Quoted in #112

src/tdastro/astro_utils/passbands.py
mi-dai: I think we've decided to not do conversions at this stage. The problem is the LSST filters online are in "nm", so we would have to do the conversion offline instead of downloading it directly

OliviaLynn: Referenced in #113

src/tdastro/astro_utils/passbands.py
OliviaLynn: Referenced in #113

src/tdastro/astro_utils/passbands.py
```diff
@@ -56,44 +77,92 @@ def __str__(self) -> str:
...
         bandfluxes = {}
-        for label, passband in self.passbands.items():
-            bandfluxes[label] = passband.fluxes_to_bandflux(flux_density_matrix, wavelengths_angstrom)
+        for full_name, passband in self.passbands.items():
+            # We only want the fluxes that are in the passband's wavelength range
+            # So, find the indices in the group's wave grid that are in the passband's wave grid
+            lower, upper = passband.waves[0], passband.waves[-1]
+            lower_index, upper_index = np.searchsorted(self.waves, [lower, upper])
+            # Check that this is the right grid after all (check will fail if the grid is not regular, or
+            # passbands are overlapping)
```
mi-dai, line 156: I'm a little confused why the check would fail in these scenarios.

mi-dai:
Sorry my review is late. After looking at the code, I would like to keep advocating for the following interface: 
```source = Source(...)
passband = Passband(delta_wave: float | None = 5, trim_percentile: float | None = 0.1)
band_flux = source.get_band_fluxes(passband, t)
```
This will avoid confusion from users having to call
```
spectral_flux = source.evaluate(passband.waves, t)
```
and having to know that the `evaluate` should be on the `passband.waves` grid.
And not having to check if user has actually call `evaluate` with the correct grid.

src/tdastro/astro_utils/passbands.py
OliviaLynn: **Units**: I do agree--but for now I did preserve the code Micheal wrote, feeling that refactoring it to be out of scope of the goal of this PR.  I've recorded details for unit changes we want to make in this issue: #117

**Check:** In most cases, we shouldn't need this; however, if the user supplies transmission tables that overlap and do not share the same step (and the user specifies delta_wave=None), the values from each would be interleaved. 

eg, (and going into this level of detail just to kind of think out loud/review this for myself here)
imagine a group with two passbands:
`passband_a.waves = [2, 4, 6]`
`passband_b.waves = [5, 6, 7, 8]` 
which results in:
`group.waves = [2, 4, 5, 6, 7, 8]`
and if we just grabbed waves based on lower and upper bounds, we would get end up grabbing that extra "5" value when generating our indices:
`[2, 4, 5, 6, 7, 8]`
`[1, 1, 1, 1, 0, 0]` <-- mask representing the indices we get
when what we really want is:
`[2, 4, 5, 6, 7, 8]`
`[1, 1, 0, 1, 0, 0]` <--excluding the 5
It's a small edge case, but possible to encounter under the specs I've been given (my understanding from the last time I spoke to @hombit Kostya is that we would like to support non-uniform transmission tables, but I asked him a while ago and fairly informally, so it could be worth revisiting). 
The check is at worse O(n), but if we can guarantee uniform table steps we could get rid of this entirely.

mi-dai: Thanks! That makes sense to me now. I think my confusion is does the check fails on one of the conditions or it only fails on both conditions (non-uniform grid AND overlapping). My understanding is the latter.

mi-dai: (but the doc implies the former?)

PR #114, opened 2024-09-18
fix snia test

mi-dai:


OliviaLynn (approved):
That makes so much more sense! Thank you! You're a genius!

Tagging you in #116, as I realize the unit tests are failing here for other reasons and masking what this fix is doing!

PR #119, opened 2024-09-23
Provide ability to set whether we compute gradients at parameters

jeremykubica:
Allow us to set whether we compute a gradient for a given parameter in the PyTree. The default is `True` for all constant and `False` for everything else. But we can now do overrides.

Also removes the `required` and `fixed` fields from the setter info as these are not needed. The `fixed` setting has been replaced by `allow_gradient`. The `required` field is not needed because the program will still fail during sampling when required data is missing. Before it was only doing a check during graph construction on some of the input paths. This will make the behavior consistent across different paths.

### DIFF
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -279,9 +279,9 @@ class Passband:
 
         # Correct for units
         if self.units == "nm":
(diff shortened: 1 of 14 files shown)


hombit (approved):
Looks goo, thank you!

PR #120, opened 2024-09-24
Passband optimization: store wave indices in each Passband

### LINKED ISSUE #112: Passband optimization
When calling `fluxes_to_bandfluxes`, we recalculate the portion of `PassbandGroup.waves` that overlaps with each individual `Passband`'s `waves`. This could be done ahead of time and stored to make `fluxes_to_bandfluxes` faster (and recalculated each time the transmission table is re-gridded).

> This looks like it isn't unique per function call. If so, it would be more efficient to do these operations once when setting up the data structure. That would mean keeping a separate dictionary of full_name to (lower_index, upper_index), so you'd need to look at how much time it saves for the added complexity.

_Originally posted by @jeremykubica in https://github.com/lincc-frameworks/tdastro/pull/111#discussion_r1763146109_

> Again, I think this could be precomputed during data creation and stored in a dictionary mapping full_name to a boolean indicating that the arrays are equal.

_Originally posted by @jeremykubica in https://github.com/lincc-frameworks/tdastro/pull/111#discussion_r1763147523_

Also, see this [comment](https://github.com/lincc-frameworks/tdastro/pull/111#discussion_r1765559192) explaining the check just following the calculation (`if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):`), quoted here:

> In most cases, we shouldn't need this; however, if the user supplies transmission tables that overlap and do not share the same step (and the user specifies delta_wave=None), the values from each would be interleaved. 

> eg, imagine a group with two passbands:
`passband_a.waves = [2, 4, 6]`
`passband_b.waves = [5, 6, 7, 8]` 
which results in:
`group.waves = [2, 4, 5, 6, 7, 8]`
and if we just grabbed waves based on lower and upper bounds, we would get end up grabbing that extra "5" value when generating our indices:
`[2, 4, 5, 6, 7, 8]`
`[1, 1, 1, 1, 0, 0]` <-- mask representing the indices we get
when what we really want is:
`[2, 4, 5, 6, 7, 8]`
`[1, 1, 0, 1, 0, 0]` <--excluding the 5
It's a small edge case, but possible to encounter under the specs I've been given (my understanding from the last time I spoke to Kostya is that we would like to support non-uniform transmission tables, but I asked him a while ago and fairly informally, so it could be worth revisiting). 
The check is at worse O(n), but if we can guarantee uniform table steps we could get rid of this entirely.

OliviaLynn:
Described in and closes #112:

> When calling `fluxes_to_bandfluxes`, we recalculate the portion of `PassbandGroup.waves` that overlaps with each individual `Passband`'s `waves`. This could be done ahead of time and stored to make `fluxes_to_bandfluxes` faster (and recalculated each time the transmission table is re-gridded).

### DIFF
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -108,9 +108,11 @@ class PassbandGroup:
             self.waves = np.unique(np.concatenate([passband.waves for passband in self.passbands.values()]))
 
     def _calculate_in_band_wave_indices(self) -> None:
-        """Calculate the indices of the groupd's wave grid that are in the passband's wave grid.
+        """Calculate the indices of the group's wave grid that are in the passband's wave grid.
+        
+        Eg, if a group's waves are [11, 12, 13, 14, 15] and a single band's are [13, 14], we get [2, 3].
 
-        The indices are stored in the passband's in_band_wave_indices attribute as either a tuple of two ints
+        The indices are stored in the passband's _in_band_wave_indices attribute as either a tuple of two ints
         (lower, upper) or a 1D np.ndarray of ints.
         """
         for passband in self.passbands.values():
@@ -121,7 +123,7 @@ class PassbandGroup:
 
             # Check that this is the right grid after all (check will fail if passbands overlap and passbands
             # do not happen to be on the same phase of the grid; eg, even if the step is 10, if the first
-            # passband starts at 100 and the second at 105, the second passband will share the same grid)
+            # passband starts at 100 and the second at 105, the passbands won't share the same grid)
             if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):
                 indices = (lower_index, upper_index + 1)
             else:
@@ -175,7 +177,7 @@ class PassbandGroup:
 
             if indices is None:
                 raise ValueError(
-                    f"Passband {full_name} does not have in_band_wave_indices set. "
+                    f"Passband {full_name} does not have _in_band_wave_indices set. "
                     "This should have been calculated in PassbandGroup._calculate_in_band_wave_indices."
                 )
 
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -109,7 +109,7 @@ class PassbandGroup:
 
     def _calculate_in_band_wave_indices(self) -> None:
         """Calculate the indices of the group's wave grid that are in the passband's wave grid.
-        
+
         Eg, if a group's waves are [11, 12, 13, 14, 15] and a single band's are [13, 14], we get [2, 3].
 
         The indices are stored in the passband's _in_band_wave_indices attribute as either a tuple of two ints
@@ -127,7 +127,7 @@ class PassbandGroup:
             if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):
                 indices = (lower_index, upper_index + 1)
             else:
-                indices = np.unique(np.searchsorted(self.waves, passband.waves))
+                indices = np.searchsorted(self.waves, passband.waves)
             passband._in_band_wave_indices = indices
 
     def process_transmission_tables(
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -69,6 +69,7 @@ class PassbandGroup:
                 self.passbands[passband.full_name] = passband
 
         self._update_waves()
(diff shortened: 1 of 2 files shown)


src/tdastro/astro_utils/passbands.py
```diff
@@ -106,6 +107,27 @@ def _update_waves(self) -> None:
...
+        """
+        for passband in self.passbands.values():
+            # We only want the fluxes that are in the passband's wavelength range
+            # So, find the indices in the group's wave grid that are in the passband's wave grid
+            lower, upper = passband.waves[0], passband.waves[-1]
+            lower_index, upper_index = np.searchsorted(self.waves, [lower, upper])
+
+            # Check that this is the right grid after all (check will fail if passbands overlap and passbands
+            # do not happen to be on the same phase of the grid; eg, even if the step is 10, if the first
+            # passband starts at 100 and the second at 105, the second passband will share the same grid)
```
OliviaLynn, line 124:
```suggestion
            # passband starts at 100 and the second at 105, the passband won't share the same grid)
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -106,6 +107,27 @@ def _update_waves(self) -> None:
         else:
             self.waves = np.unique(np.concatenate([passband.waves for passband in self.passbands.values()]))
 
+    def _calculate_in_band_wave_indices(self) -> None:
+        """Calculate the indices of the groupd's wave grid that are in the passband's wave grid.
```
OliviaLynn, line 111:
```suggestion
        """Calculate the indices of the group's wave grid that are in the passband's wave grid.
        
        Eg, if a group's waves are [11, 12, 13, 14, 15] and a single band's are [13, 14], we get [2, 3].
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -106,6 +107,27 @@ def _update_waves(self) -> None:
...
+
+        The indices are stored in the passband's in_band_wave_indices attribute as either a tuple of two ints
```
OliviaLynn, line 113:
```suggestion
        The indices are stored in the passband's _in_band_wave_indices attribute as either a tuple of two ints
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -148,18 +171,17 @@ def fluxes_to_bandfluxes(self, flux_density_matrix: np.ndarray) -> np.ndarray:
...
-            lower_index, upper_index = np.searchsorted(self.waves, [lower, upper])
-            # Check that this is the right grid after all (check will fail if the grid is not regular, or
-            # passbands are overlapping)
-            if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):
-                in_band_fluxes = flux_density_matrix[:, lower_index : upper_index + 1]
+            indices = passband._in_band_wave_indices
+
+            if indices is None:
+                raise ValueError(
+                    f"Passband {full_name} does not have in_band_wave_indices set. "
```
OliviaLynn, line 178:
```suggestion
                    f"Passband {full_name} does not have _in_band_wave_indices set. "
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -106,6 +107,29 @@ def _update_waves(self) -> None:
...
+            lower, upper = passband.waves[0], passband.waves[-1]
+            lower_index, upper_index = np.searchsorted(self.waves, [lower, upper])
+
+            # Check that this is the right grid after all (check will fail if passbands overlap and passbands
+            # do not happen to be on the same phase of the grid; eg, even if the step is 10, if the first
+            # passband starts at 100 and the second at 105, the passbands won't share the same grid)
+            if np.array_equal(self.waves[lower_index : upper_index + 1], passband.waves):
+                indices = (lower_index, upper_index + 1)
+            else:
+                indices = np.searchsorted(self.waves, passband.waves)
```
jeremykubica, line 130: In the original code, you had a `np.unique` call here. Is that no longer needed?

jeremykubica (approved):

src/tdastro/astro_utils/passbands.py
OliviaLynn: Yes--the non-unique values I had been seeing were a result of me having gotten the arguments to `np.searchsorted` flipped (was previously `(passband.waves, self.waves)`).

But as long as all elements of `passband.waves` are included in `self.waves`, and neither of them contain duplicate/non-unique values, all elements of the returned array should already be unique.

PR #122, opened 2024-09-24
Move redshift from effect to physical model

### LINKED ISSUE #76: Move redshift to Physical Model

OliviaLynn:
Described in and closes #76 (pardon the mismatch in the branch name):

> Instead of applying redshift as an effect, I think we should make it part of PhysicalModel. If redshift != 0, PhysicalModel.evaluate always calls the pre-effect function to transform the frame and then always calls the post effect afterward. This way the user does not need to manually add redshift effects.

### DIFF
--- a/src/tdastro/sources/physical_model.py
+++ b/src/tdastro/sources/physical_model.py
@@ -116,6 +116,49 @@ class PhysicalModel(ParameterizedNode):
         for effect in self.effects:
             effect.set_graph_positions(seen_nodes=seen_nodes)
 
+    def _redshift_pre_effect(self, observer_frame_times, observer_frame_wavelengths, params):
+        """Calculate the rest-frame times and wavelengths needed to give us the observer-frame times
+        and wavelengths (given the redshift).
+
+        Parameters
+        ----------
+        observer_frame_times : numpy.ndarray
+            The times at which the observation is made.
+        observer_frame_wavelengths : numpy.ndarray
+            The wavelengths at which the observation is made (in angstroms).
+        params : `dict`
+            A dictionary of the model's parameters.
+
+        Returns
+        -------
+        tuple of (numpy.ndarray, numpy.ndarray)
+            The rest-frame times and wavelengths needed to generate the rest-frame flux densities,
+            which will later be redshifted  back to observer-frame flux densities at the observer-frame
+            times and wavelengths.
+        """
+        observed_times_rel_to_t0 = observer_frame_times - params["t0"]
+        rest_frame_times_rel_to_t0 = observed_times_rel_to_t0 / (1 + params["redshift"])
+        rest_frame_times = rest_frame_times_rel_to_t0 + params["t0"]
+        rest_frame_wavelengths = observer_frame_wavelengths / (1 + params["redshift"])
+        return (rest_frame_times, rest_frame_wavelengths)
+
+    def _redshift_post_effect(self, flux_density, params):
+        """Apply the redshift effect to observations (rest-frame flux_density values).
+
+        Parameters
+        ----------
+        flux_density : `numpy.ndarray`
+            A length T X N matrix of flux density values (in nJy).
+        params : `dict`
+            A dictionary of the model's parameters.
+
+        Returns
+        -------
+        flux_density : `numpy.ndarray`
+            The redshifted results (in nJy).
+        """
+        return flux_density / (1 + params["redshift"])
+
     def _evaluate(self, times, wavelengths, graph_state):
         """Draw effect-free observations for this object.
 
@@ -172,9 +215,8 @@ class PhysicalModel(ParameterizedNode):
         params = self.get_local_params(graph_state)
 
         # Pre-effects are adjustments done to times and/or wavelengths, before flux density computation.
-        for effect in self.effects:
-            if hasattr(effect, "pre_effect"):
-                times, wavelengths = effect.pre_effect(times, wavelengths, graph_state, **kwargs)
+        if self.redshift is not None and self.redshift != 0.0:
+            times, wavelengths = self._redshift_pre_effect(times, wavelengths, params)
 
         # Compute the flux density for both the current object and add in anything
         # behind it, such as a host galaxy.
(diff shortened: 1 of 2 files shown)


src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
         for effect in self.effects:
             effect.set_graph_positions(seen_nodes=seen_nodes)
 
+    def _redshift_pre_effect(self, observer_frame_times, observer_frame_wavelengths, params):
+        """Calculate the rest-frame times and wavelengths needed to give us the observer-frame times
```
OliviaLynn, line 120:
```suggestion
        """Calculate the rest frame times and wavelengths needed to give user the observer frame times
```

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+        and wavelengths (given the redshift).
+
+        Parameters
+        ----------
+        observer_frame_times : numpy.ndarray
+            The times at which the observation is made.
+        observer_frame_wavelengths : numpy.ndarray
+            The wavelengths at which the observation is made (in angstroms).
```
OliviaLynn, line 128:
```suggestion
            The wavelengths at which the observation is made (in Angstroms).
```

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+        params : `dict`
+            A dictionary of the model's parameters.
+
+        Returns
+        -------
+        tuple of (numpy.ndarray, numpy.ndarray)
+            The rest-frame times and wavelengths needed to generate the rest-frame flux densities,
```
OliviaLynn, line 135:
```suggestion
            The rest frame times and wavelengths needed to generate the rest frame flux densities,
```

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+            which will later be redshifted  back to observer-frame flux densities at the observer-frame
```
OliviaLynn, line 136:
```suggestion
            which will later be redshifted  back to observer frame flux densities at the observer frame
```

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+            A dictionary of the model's parameters.
+
+        Returns
+        -------
+        tuple of (numpy.ndarray, numpy.ndarray)
+            The rest frame times and wavelengths needed to generate the rest frame flux densities,
+            which will later be redshifted  back to observer frame flux densities at the observer frame
+            times and wavelengths.
+        """
+        observed_times_rel_to_t0 = observer_frame_times - params["t0"]
```
jeremykubica, line 139: What happens if t0 is not set? We have a few options. We could skip the computation, fail a check, or use a reasonable default. How about something like:
```
t0 = params.get("t0", 0.0)
observed_times_rel_to_t0 = observer_frame_times - t0
```
or if we change how the parameters are passed (using `t0` and `redshift` instead of `params`) we could just make it:
```
observed_times_rel_to_t0 = observer_frame_times - t0
```

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+        """Calculate the rest frame times and wavelengths needed to give user the observer frame times
+        and wavelengths (given the redshift).
+
+        Parameters
+        ----------
+        observer_frame_times : numpy.ndarray
+            The times at which the observation is made.
+        observer_frame_wavelengths : numpy.ndarray
+            The wavelengths at which the observation is made (in Angstroms).
+        params : `dict`
```
jeremykubica, line 129: I have a slight preference for passing the redshift and t0 explicitly so this function doesn't need to worry about how the dictionary is stored. The reason is that it lets users know exactly what is expected by the function.

If you'd prefer to keep it as-is (passing in `params`), you should add a note to the doc string that params needs to include redshift and t0.

src/tdastro/sources/physical_model.py
```diff
@@ -116,6 +116,49 @@ def set_graph_positions(self, seen_nodes=None):
...
+            which will later be redshifted  back to observer frame flux densities at the observer frame
+            times and wavelengths.
+        """
+        observed_times_rel_to_t0 = observer_frame_times - params["t0"]
+        rest_frame_times_rel_to_t0 = observed_times_rel_to_t0 / (1 + params["redshift"])
+        rest_frame_times = rest_frame_times_rel_to_t0 + params["t0"]
+        rest_frame_wavelengths = observer_frame_wavelengths / (1 + params["redshift"])
+        return (rest_frame_times, rest_frame_wavelengths)
+
+    def _redshift_post_effect(self, flux_density, params):
```
jeremykubica, line 145: As above, I think it is more understandable if we just pass in `redshift` directly instead of the dictionary. We can let earlier functions unpack the dictionary as needed.

tests/tdastro/effects/test_redshift.py
```diff
@@ -1,71 +0,0 @@
...
-
-    # Check shape of output is as expected
-    assert values_no_effects.shape == (len(times), len(wavelengths))
-    assert values_redshift.shape == (len(times), len(wavelengths))
-
-    return values_no_effects, values_redshift
-
-
-def test_redshift() -> None:
-    """Test that we can create a Redshift object and it gives us values as expected."""
```
jeremykubica, line 23: Is it worth preserving this test? I'm guessing not, because the `test_redshift_values()` looks more comprehensive. But I wanted to check.

src/tdastro/sources/physical_model.py
OliviaLynn: Now raising ValueError when redshift is set but t0 is not (checking this in PhysicalModel.evaluate)

src/tdastro/sources/physical_model.py
OliviaLynn: Now directly passing redshift and t0 to `obs_frame_to_rest_frame` and redshift to `apply_redshift`

tests/tdastro/effects/test_redshift.py
OliviaLynn: It felt redundant to me--I think I needed this when I was initially figuring out redshift, but I wouldn't say this offers anything novel (and so is misleading to include it).

jeremykubica (approved):

PR #126, opened 2024-09-25
Remove the EffectModel class

jeremykubica:
Instead of having a arbitrary list of effects for the parameterized node, we are going to support a few productionized effects directly in parts of the code such as `PhysicalModel`. This will reduce complexity of the code and make it more difficult to make ordering errors.


OliviaLynn (approved):
Looks good!

PR #130, opened 2024-09-26
Create a proof of concept JAX implementation of SALT2

jeremykubica:
This is a minimal implementation of SALT2 using JAX. It is currently a lot slower than the sncosmo version (and can likely be optimized), but produces the correct results.

The PR includes a JAX version of sncosmo's bicubic interpolator (again this is slower than the cython version).


src/tdastro/utils/bicubic_interp.py
```diff
@@ -0,0 +1,242 @@
...
+"""
+
+import jax.numpy as jnp
+from jax import jit, vmap
+from jax.lax import cond
+
+from tdastro.utils.io_utils import read_grid_data
+
+
+@jit
```
hombit, line 14: I'm not sure that we need `@jit` on this level, should we compile the code using cubic interpolation instead?

tests/tdastro/sources/test_salt2.py
hombit, line 1: I'd test it vs sncosmo

hombit (commented):
Looks good, but I'd ask for more testing vs implementations in other packages

tests/tdastro/sources/test_salt2.py
jeremykubica: I did that during development, but I want to move away from tests that need to download sncosmo models. The download unpredictability is causing intermittent failures already. We should try to make the tests as hermetic as possible.

hombit: Could we still run both with the test data you prepared?

src/tdastro/sources/salt2_jax.py
```diff
@@ -0,0 +1,120 @@
...
+        # Use the default color correction values.
+        self._colorlaw = SALT2ColorLaw.from_file(os.path.join(model_dir, cl_filename))
+
+    def _evaluate(self, phase, wavelengths, graph_state, **kwargs):
+        """Draw effect-free observations for this object.
+
+        Parameters
+        ----------
+        phase : `numpy.ndarray`
+            A length T array of rest frame timestamps.
```
mi-dai, line 98: This is the rest frame phases (which is timestamps - time_at_peak). 
We need to add the t0 parameter like in https://github.com/lincc-frameworks/tdastro/blob/main/src/tdastro/sources/sncomso_models.py#L32
so `_evaluate()` is consistent with other physical models as `_evaluate(self, times, wavelengths ...)`

jeremykubica:
Closing because we don't need this at the moment.

Issue #127, opened 2024-09-25
LSST g band file is wrong


hombit:
**Bug report**

LSST g band file looks different from others: wavelength range is different from others and all wavelengths are 4-digit, like they are in angstroms instead of being in nm.

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

hombit:
We should use "total" transmissions, for example this file should be used for g-band
https://github.com/lsst/throughputs/blob/22732801ff32be387de00b5888ccd3e0c787f422/baseline/total_g.dat

So we should also update other files to use total transmissions. Also, it would help to track it if we put this commit-specific links to the passbands README file

OliviaLynn:
Thanks for bringing this to my attention! I must have copied/pasted the files over before, and there was some mismatch between the files from the old url and the new url.

I went ahead and replaced all the files (with the current url in use) in the test dir as part of #135

hombit:
@hombit sounds great! Could you also check that other bands are from "total_*" files, not "filter_*"?

OliviaLynn:
Checked and they are!

Issue #113, opened 2024-09-18
Passband general improvements


OliviaLynn:
- [ ] We now have LSST transmission tables stored in our test data directory (and a `passband_dir` fixture defined in our conftest); it would be nice to rewrite some of our unit tests to use these tables rather than toy data

- [ ] Use astropy units in Passband code
> Should we just use astropy units here so we can use all their conversions?

_Originally posted by @jeremykubica in https://github.com/lincc-frameworks/tdastro/pull/111#discussion_r1763151241_

- [ ] Refactor Passband constructor
> I think it is fine to keep this as-is, but an alternative approach that might be more modular approach might be to have the constructor take a table and then to add two class methods:

```
@classmethod
def from_file(cls, table_path: str, ...):
   load the file
   ...
   return Passband(table, ...)

@classmethod
def from_url(cls, table_url: str, force_download: bool, ...):
    do the download
    ...
    return Passband(table, ...)
```

_Originally posted by @jeremykubica in https://github.com/lincc-frameworks/tdastro/pull/111#discussion_r1763156472_

PR #135, opened 2024-09-26
Use LSST transmission test data for passband unit tests

OliviaLynn:
For the first part of #113 (updating unit tests to use LSST transmission tables stored in unit test data).

Should also cover #127 (LSST g band file is wrong).

Introduces a `table_dir` option to PassbandGroup, so you can specify a general transmission table directory to use, as in: 
- `PassbandGroup(preset="LSST", table_dir="my/dir")` 
- or to specify just a root dir when manually defining passbands with `PassbandGroup(table_dir="my/dir", passband_parameters=[...])`)


src/tdastro/astro_utils/passbands.py
```diff
@@ -48,26 +51,37 @@ def __init__(
             - units : str (either 'nm' or 'A')
             If survey is not LSST (or other survey with defined defaults), either a table_path or table_url
             must be provided.
+        table_dir : str, optional
+            The path to the directory containing the passband tables. If a table_path has not been specified
+            in the passband_parameters dictionary, table paths will be set to
+            {table_dir}/{survey}/{filter_name}.dat. If None, the table path will be set to a default path.
```
jeremykubica, line 57: No change needed right now, but I had been thinking of this as "{table_dir}/{filter_name}.dat" so the user could effectively provide both the base directory and the survey name with the table_dir parameter.

This current approach better matches the presets below though.

src/tdastro/astro_utils/passbands.py
```diff
@@ -81,25 +95,31 @@ def __str__(self) -> str:
...
-            }
+            for filter_name in ["u", "g", "r", "i", "z", "y"]:
+                if table_dir is None:
+                    self.passbands[f"LSST_{filter_name}"] = Passband("LSST", filter_name, **kwargs)
+                else:
+                    table_path = os.path.join(table_dir, "LSST", f"{filter_name}.dat")
+                    self.passbands[f"LSST_{filter_name}"] = Passband(
+                        "LSST",
+                        filter_name,
+                        **{**kwargs, "table_path": table_path},
```
jeremykubica, line 121: I think it would be more readable to just do:
```
table_path=table_path,
**kwargs,
)
```

tests/tdastro/astro_utils/test_passband_groups.py
```diff
@@ -152,15 +136,59 @@ def test_passband_group_wrapped_from_physical_source(tmp_path):
...
+    evaluated_fluxes = model.evaluate(test_times, toy_passband_group.waves, state)
+    result_from_passband_group = toy_passband_group.fluxes_to_bandfluxes(evaluated_fluxes)
 
     # Check the two dicts are the same
     assert result_from_source_model.keys() == result_from_passband_group.keys()
     for key in result_from_source_model:
         np.testing.assert_allclose(result_from_source_model[key], result_from_passband_group[key])
+
+
+def test_passband_group_calculate_in_band_wave_indices(passbands_dir, tmp_path):
```
hombit, line 163: I'm actually a bit confused about the original function and different return formats for different cases. If a passband waves are represented by a range of `catalog.waves` it returns `(lower_index, upper_index + 1)` tuple, while for the opposite case it is an integer array of indexes. Than here it is used to index an array:
https://github.com/lincc-frameworks/tdastro/blob/7450fccec5d30c5737d2d21772ac8d7421315c12/src/tdastro/astro_utils/passbands.py#L187-L190

I'd suggest to return `slice(lower_index, upper_index + 1)` instead, than both types of the return indexes may be just used directly.

Should I make a separate issue for that?

OliviaLynn: Added

PR #140, opened 2024-09-30
Fix the HostmassX1Func return type

### LINKED ISSUE #134: Inconsistency in SN Ia utility functions return values
**Bug report**

Some of SN Ia utility functions return float, while some others return a single-element 1-d array. So in the SN Ia end2end test, `source.get_param(state, parname)` returns `float` for `parname` "t0", "c", "redshift", "ra", "dec" and single-element 1-d array for "x1", "x2".
https://github.com/lincc-frameworks/tdastro/blob/bcd50785bc63527bc8769a51f4eff17e0701f051/tests/tdastro/sources/test_snia.py#L167

**Before submitting**
Please check the following:

- [ ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.

jeremykubica:
Closes #134 

`HostmassX1Func` should return a float (instead of a numpy array with one element) when `num_samples` is 1.

### DIFF
--- a/src/tdastro/astro_utils/snia_utils.py
+++ b/src/tdastro/astro_utils/snia_utils.py
@@ -251,7 +251,9 @@ class HostmassX1Func(NumericalInversePolynomialFunc):
         hostmass = self.get_param(graph_state, "hostmass")
 
         if graph_state.num_samples == 1:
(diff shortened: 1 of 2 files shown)


hombit (approved):
Looks good! Do i understand correctly that it also should fix it for x0?

jeremykubica:
Correct. It will fix it for x0 because that was taking x1 as an array input.

PR #143, opened 2024-10-01
Update ASV scripts

jeremykubica:
Reapply the python project template (`copier update --vcs-ref=HEAD`) to update the ASV scripts.


hombit (approved):
Thank you

PR #148, opened 2024-10-02
Adopt nJy in blackbody model and eclipsing binaries

hombit:
This fixes periodic model to return values in nJy, and adds some comments to black-body model to highlight that it has CGS inputs and outputs.

### DIFF
--- a/src/tdastro/consts.py
+++ b/src/tdastro/consts.py
@@ -20,3 +20,10 @@ This is derived from two facts for a symmetric 2D Gaussian:
    where
    g(x,y) = 1 / (2π sigma²) exp(-(x²+y²)/2/sigma²).
 """
+
+ANGSTROM_TO_CM = (1.0 * u.AA).to_value(u.cm)
+"""Conversion factor from Angstrom to cm, 1e-8"""
+
+
+CGS_FNU_UNIT_TO_NJY = (1.0 * u.erg / u.second / u.cm**2 / u.Hz).to_value(u.nJy)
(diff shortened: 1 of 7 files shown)


src/tdastro/consts.py
```diff
@@ -20,3 +20,10 @@
    where
    g(x,y) = 1 / (2π sigma²) exp(-(x²+y²)/2/sigma²).
 """
+
+ANGSTROM_TO_CM = (1.0 * u.AA).to_value(u.cm)
+"""Conversion factor from Angstrom to cm, 1e-8"""
+
+
+CGS_FNU_UNIT_TO_NJY = (1.0 * u.erg / u.second / u.cm**2 / u.Hz).to_value(u.nJy)
+"""Conversion factor from erg/s/cm²/Hz to nJy, 1e23"""
```
mi-dai, line 29: Do these consts need tests?

mi-dai (commented):
Looks good! Only one minor comment

src/tdastro/consts.py
hombit: Done

mi-dai, line 29: 1e32

mi-dai (changes requested):
One more change in docstring!

src/tdastro/consts.py
hombit: Fixed!

mi-dai (approved):

PR #152, opened 2024-10-08
Rename _evaluate to compute_flux()

jeremykubica:
Rename `_evaluate()` to make it distinct from `evaluate()`.


hombit (approved):
Thank you

PR #153, opened 2024-10-09
Basic IO Functions for GraphState

jeremykubica:
Add some IO and helper functions for `GraphState`:
- Transform a `GraphState` to/from an AstroPy Table.
- Save/Read a `GraphState` to a ecsv file.
- Compare two `GraphState`s for equality.
These will be used in the future to save program state and output the lightcurves.


OliviaLynn (approved):
Looks good!

PR #158, opened 2024-10-10
Move the run_snia_end2end out of tests

jeremykubica:
Move functions out of test_snia so they can be used in both the test and notebook.


hombit (approved):
Thank you!

PR #160, opened 2024-10-14
Add a bunch of helper functions

jeremykubica:
A sequence of API clean ups that will make simulating and saving light curves easier:
- If a node’s `nodel_label` parameter is set, use that as the full node name. This will allow us to have names such as “source” and “host” instead of “0:source”.
- Provide the ability to set node labels (`set_node_label()`).
- Allow `ParameterizedNode.get_param()` to return a default value to mirror the behavior of the dictionary’s `get()` function.
- Add the ability for `GraphState` to access values from either a (node name, param name) pair or the extended name string.
- Add a `__contains__` function to `GraphState`.
- Add a function to export a `GraphState` to a dict which will allow us to create a nested-pandas data frame.


src/tdastro/graph_state.py
```diff
@@ -60,6 +60,17 @@ def __init__(self, num_samples=1):
...
         return self.num_parameters
 
+    def __contains__(self, key):
+        if key in self.states:
+            return True
+        elif self._NAME_SEPARATOR in key:
+            tokens = key.split(self._NAME_SEPARATOR)
+            if len(tokens) != 2:
+                raise KeyError(f"Invalid GraphState key: {key}")
+            return True
```
OliviaLynn, line 70: Do we still want to check if self.states contains the tokens?

OliviaLynn (approved):
Looks great! Just the one comment

src/tdastro/graph_state.py
jeremykubica: Nice catch. This is a bug. I'm pushing a change to fix it (and corresponding tests).

PR #161, opened 2024-10-14
Simplify random number generation

jeremykubica:
Simplify the structure and the seeding of the random number generation. There are multiple ways to see the random number generators and multiple random number generators to use, potentially causing confusion.

This PR standardizes the code to use numpy's random number generators. This is chosen over JAX because: 1) it has more understandable semantics, 2) early users always defaulted to numpy anyway, 3) we should not actually need JAX random number generation for the auto-differentiation (those parameters will be sampled ahead of time), and 4) the vectorized numpy operations are already pretty fast. We can always add it back later if needed. But standardizing for now greatly simplifies the interface.

In addition this PR removes the ability to pass a dictionary of random number generators to the nodes (such that each node gets its own pre-defined random number generator). Instead the user can:

1) Use the node's default random number generator for truly random results. Note that we can still debug and reproduce results in this case because we have the ability to snapshot and save the `GraphState` object. So we can playback the exact computation ay any point.

2) Use a given per-node seed for strict testing and reproducibility. Do not use this for generating scientific results.

3) Pass a single random number generator to be used for the entire graph (for testing and reproducibility).

Once we are no longer building dictionaries of random number generators, we can remove the helper functions (e.g. `build_rngs_from_hashes`) and the `ParameterizedModel. get_all_node_info()` function.


src/tdastro/astro_utils/snia_utils.py
```diff
@@ -235,9 +235,9 @@ def compute(self, graph_state, rng_info=None, **kwargs):
         graph_state : `GraphState`
             An object mapping graph parameters to their values. This object is modified
             in place as it is sampled.
-        rng_info : `dict`, optional
-            A dictionary of random number generator information for each node, such as
-            the JAX keys or the numpy rngs.
+        rng_info : numpy.random._generator.Generator, optional
```
hombit, line 238: Should it be? Or people'd understand that `optional` usually mean that it can be `None`?
```suggestion
        rng_info : numpy.random._generator.Generator or None, optional
```

hombit (approved):
I do enjoy these changes, thank you! This should help debugging models and pipelines.

I'm not fully understand how parallel sampling will work, but I think we will see it in the future code examples.

PR #164, opened 2024-10-15
Remove flush that is causing benchmarking errors

jeremykubica:
I ran into a problem with `flush` causing an error while writing benchmarks. I need to commit this fix before the new benchmarks.


hombit (approved):
TIL

PR #165, opened 2024-10-15
Expand the benchmark suite

jeremykubica:
Add a lot more benchmarks based on the snia_test notebook.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -5,29 +5,155 @@ To manually run the benchmarks use: asv run
 For more information on writing benchmarks:
 https://asv.readthedocs.io/en/stable/writing_benchmarks.html."""
 
-from tdastro.astro_utils.snia_utils import HostmassX1Func
+import os
+
+import numpy as np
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks.py
```diff
@@ -5,29 +5,155 @@
...
+
+
+def _load_test_passbands():
+    """Load passbands to use in various benchmarks."""
+    passbands_dir = os.path.join(tdastro._TDASTRO_TEST_DATA_DIR, "passbands")
+    passbands = PassbandGroup(
+        passband_parameters=[
+            {
+                "filter_name": "g",
+                "table_path": f"{passbands_dir}/LSST/g.dat",
```
hombit, line 28: f-string path concatenation!!!!111
CC @delucchi-cmu

hombit (approved):
This is very nice, thank you

benchmarks/benchmarks.py
jeremykubica: Fixed

PR #169, opened 2024-10-15
Allow NumericalInversePolynomialFunc to take a function

### LINKED ISSUE #166: Make a distribution from a (log-)PDF function
**Feature request**

It would be nice to have a couple of thin wrappers around `NumericalInversePolynomialFunc` to provide a convenient way to wrap UDF (log-) PDF function into a distribution object.

Pseudo-implementation:

```python
class PDF:
    def __init__(self, f):
        self.pdf = f

class LogPDF:
    ...

def distribution_from_pdf(f):
    return NumericalInversePolynomialFunc(PDF(f))

def distribution_from_logpdf(f):
    return NumericalInversePolynomialFunc(LogPDF(f))
```

**Before submitting**
Please check the following:

jeremykubica:
Closes #166 

Allows the user to give `NumericalInversePolynomialFunc` a PDF function directly instead of wrapping it in a class. Includes an example distribution and calls in the test.

### DIFF
--- a/src/tdastro/rand_nodes/scipy_random.py
+++ b/src/tdastro/rand_nodes/scipy_random.py
@@ -9,15 +9,29 @@ from tdastro.base_models import FunctionNode
 from tdastro.graph_state import transpose_dict_of_list
 
 
+class PDFFunctionWrapper:
+    """A class that just wraps a given PDF function.
+
+    Attributes
+    ----------
+    _pdf : function
+        The PDF function.
+    """
+
+    def __init__(self, func):
+        self._pdf = func
+        self.pdf = self._pdf
+
+
 class NumericalInversePolynomialFunc(FunctionNode):
     """A class for sampling from scipy's NumericalInversePolynomial
-    given an distribution object or a class from which to create
-    such an object.
+    given a distribution function, an object with a pdf function,
+    or a class from which to create such an object.
 
     Note
     ----
     If a class is provided, then the sampling function will create a new
-    object (with the sampled parameters) for each sampling.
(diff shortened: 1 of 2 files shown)


src/tdastro/rand_nodes/scipy_random.py
```diff
@@ -9,15 +9,29 @@
...
+    Attributes
+    ----------
+    _pdf : function
+        The PDF function.
+    """
+
+    def __init__(self, func):
+        self._pdf = func
+        self.pdf = self._pdf
+
```
mi-dai, line 24: Can we add `logpdf` here as well?

jeremykubica: Sure. I added this. It changes how the overloading works. We can't infer pdf vs logpdf from the function alone (at least not quickly), so I created two subclasses that handle each of these cases.

src/tdastro/rand_nodes/scipy_random.py
```diff
@@ -162,3 +166,69 @@ def generate(self, given_args=None, num_samples=1, rng_info=None, **kwargs):
...
+        The log PDF function.
+    """
+
+    def __init__(self, func):
+        self._log_pdf = func
+        self.logpdf = self._log_pdf
+
+
+class SampleLogPDF(NumericalInversePolynomialFunc):
+    """A node for sampling from a given PDF function.
```
mi-dai, line 219:
```suggestion
    """A node for sampling from a given Log PDF function.
```

mi-dai (approved):

PR #170, opened 2024-10-15
Add a helper function for looking at variables

jeremykubica:
For common parameters, we might want to extract them from the GraphState without going through the node. This is useful in both logging and debugging. This PR adds a helper function that can extract (uniquely named) parameters from arbitrary nodes in the graph.

### DIFF
--- a/src/tdastro/graph_state.py
+++ b/src/tdastro/graph_state.py
@@ -321,47 +321,67 @@ class GraphState:
                     new_state.states[node_name][var_name] = value[sample_num]
         return new_state
 
-    def extract_parameters(self, params=None, nodes=None):
+    def extract_parameters(self, params):
         """Extract the parameter value(s) by a given name. This is often used for
         recording the important parameters from an entire model (set of nodes).
 
         Parameters
         ----------
         params : str or list-like, optional
-            The parameter names to extract. If None (not provided), extract data
-            for all parameters.
-            Default: None
-        nodes: str or list-like, optional
-            The node names to extract. If None (not provided), extract data
-            from all nodes.
-            Default: None
+            The parameter names to extract. These can be full names ("node.param") or
+            use the parameter names.
 
         Returns
         -------
         values : dict
             The resulting dictionary.
         """
-        if nodes is None:
-            use_nodes = None
-        elif isinstance(nodes, str):
-            use_nodes = set([nodes])
-        else:
-            use_nodes = set(nodes)
+        # If we are looking up a single parameter, but it into a list.
+        if isinstance(params, str):
+            params = [params]
 
-        if params is None:
-            use_params = None
-        elif isinstance(params, str):
-            use_params = set([params])
-        else:
-            use_params = set(params)
+        # Go through all the parameters. If a parameters fill name is provided,
+        # look it up now and save the result. Otherwise put it into a list to check
+        # for in each node.
+        single_params = set()
+        results = {}
+        for current in params:
+            if "." in current:
+                node_name, param_name = current.split(".")
+                if node_name in self.states and param_name in self.states[node_name]:
+                    results[current] = self.states[node_name][param_name]
+            else:
+                single_params.add(current)
 
-        values = {}
+        if len(single_params) == 0:
+            # Nothing else to do.
+            return results
+
+        # Traverse the trnested dictionaries looking for cases where the parameter names match.
+        first_seen_node = {}
         for node_name, node_params in self.states.items():
-            if use_nodes is None or node_name in use_nodes:
-                for param_name, param_value in node_params.items():
-                    if use_params is None or param_name in use_params:
-                        values[self.extended_param_name(node_name, param_name)] = param_value
-        return values
+            for param_name, param_value in node_params.items():
+                if param_name in single_params:
+                    if param_name in first_seen_node:
+                        # We've already seen this parameter in another node. Time to use the
+                        # expanded names.
+
+                        # Start by expanding the result we have already seen if needed.
+                        if param_name in results:
+                            full_name_existing = f"{first_seen_node[param_name]}.{param_name}"
+                            results[full_name_existing] = results[param_name]
+                            del results[param_name]
+
+                        # Add the result from the current node.
+                        full_name_current = f"{node_name}.{param_name}"
+                        results[full_name_current] = param_value
+                    else:
+                        # This is the first time we have seen the node. Save it with
+                        # just the parameter name. Also save the node where we saw it.
+                        results[param_name] = param_value
+                        first_seen_node[param_name] = node_name
+
+        return results
 
     def to_table(self):
         """Flatten the graph state to an AstroPy Table with columns for each parameter.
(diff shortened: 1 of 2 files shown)


mi-dai (commented):
After discussing with @hombit, we think we do not really need the `collapse` feature, since it's hard to define how close is close enough to be considered the same. I think we can keep just the `collapse=False` behavior.

jeremykubica:
> After discussing with @hombit, we think we do not really need the `collapse` feature, since it's hard to define how close is close enough to be considered the same. I think we can keep just the `collapse=False` behavior.

In terms of "close enough", I had been thinking effectively an exact match. If the value is exactly the same then it is likely just information copied between nodes. But if we don't need this, then it is easier not to have it.

If we do not need `collapse` then we can do this a lot easier by making filtering an option in `to_dict`. I made that change. What do you think?

mi-dai:
Looks good to me. I think we do not need `collapse` regardless of whether we can extract the exact same parameter. 
But it seems now we are not able to do `extract_parameters('node.parameter')` any more? I would still like to have that feature. Also `to_dict` doesn't sound very intuitive compared to `extract_parameters`, can we still keep `extract_parameters`?

jeremykubica:
@mi-dai I changed the name from `to_dict` to `extract_parameters`. I kept the `to_dict` function since it is being used in other contexts. It just calls `extract_parameters` now.

I also re-added the functionality to pass a single string or pass parameter names as a positional argument. I don't think we should include the full name option though, because we can already access that with the `[]` notation. I added some tests to illustrate this. We should access the value of a single parameter using the full name as `state["node_name.param_name"]` and only use `extract_parameters` to get a dictionary of multiple parameters. This saves us from having to check which type is returned.

tests/tdastro/test_graph_state.py
```diff
@@ -402,6 +402,66 @@ def test_graph_to_from_file():
...
+    assert results["c.v2"] == 5.0
+    assert results["c.v3"] == 3.0
+
+    # We can extract a set of parameters from a set of nodes.
+    results = state.extract_parameters(nodes=["a", "c", "e"], params=["v1", "v2"])
+    assert len(results) == 3
+    assert results["a.v1"] == 1.0
+    assert results["a.v2"] == 2.0
+    assert results["c.v2"] == 5.0
+
```
mi-dai, line 463: I'm thinking something like `extract_parameters(['a.v1','c.v2','v3'])` which will return {'a.v1: 1, 'c.v2': 5, 'a.v3': 3, 'c.v3': 3, 'e.v3': 3}. Or `extract_parameters(['a.v1','v5'])` which will return {'a.v1': 1, 'v5': 7} or {'a.v1': 1, 'e.v5': 7} (either is fine but maybe the former is slightly preferred).
For example, I wanted to get a quick plot of the simulated Hubble Diagram using host redshift, so I'd want to do `extract_parameters(['host.redshift','x0','x1','c'])`, and I don't care (or don't know) what node `x0`,`x1` or `c` is from.

jeremykubica: That behavior was the point of the `collapse` logic. It allowed returning a single value for a parameter using just the parameter name.

In your case of `extract_parameters(['host.redshift','x0','x1','c'])` how do you want to handle duplicates? If we find instances of `x0` in two different nodes we could:
1. Throw an error [initial behavior]
2. Collapse them if they are identical
    2a. Throw an error if they are not identical
    2b. Return both in expanded form if they are not identical [second round behavior]
3. Return all values [current behavior]
4. Return an arbitrary one of the `x0` values. [Not recommended as the behavior becomes undefined].

More concretely if someone where to call `extract_parameters(['ra', 'dec', 't0'])` where the host and source have two different values of `ra`, how would you like the code to behave?

mi-dai: I think we want to output duplicate parameter names with their node names (say `a.x0`, `b.x0`), regardless of whether they are identical in values.
If someone calls `extract_parameters(['ra','dec','t0'])`, we'll return {'host.ra','source.ra','anything_else.ra', ...}, even some of them may be identical in their values.

mi-dai: I wanted to add that I think `extract_parameters(node=[],parameters=[])` is still useful. Maybe parameters can be expended to include `extract_parameters(parameters=['host.redshift','x0','x1','c'])`?

jeremykubica: I think I've got it, but let me check.

In the case of a state:
```
a.v1 -> 1.0
a.v2 -> 2.0
a.v3 -> 3.0
b.v1 -> 4.0
b.v2 -> 5.0
b.v4 -> 6.0
c.v1 -> 7.0
```
and the query `extract_parameters(parameters=['a.v1', 'v2', 'v3', 'b.v4'])` you would want the function to return the dictionary:

```
{
    "a.v1": 1.0,
    "a.v2": 2.0,
    "b.v2": 5.0,
    "v3": 3.0,
    "b.v4": 6.0,
}
```

Specifically:
1) The full name are always preserved (even if the parameter does not exist in other nodes).
2) The shortened name is preserved if the parameter is unique.
3) The shortened name is transformed into multiple long names if the parameter is not unique.
Is that correct?

jeremykubica: I implemented those rules. I've dropped the `nodes` parameter for now, because it is not clear how to incorporate that as well. We can always extract all parameters with `to_dict`.

src/tdastro/graph_state.py
```diff
@@ -321,47 +321,67 @@ def extract_single_sample(self, sample_num):
...
+                        # Add the result from the current node.
+                        full_name_current = f"{node_name}.{param_name}"
+                        results[full_name_current] = param_value
+                    else:
+                        # This is the first time we have seen the node. Save it with
+                        # just the parameter name. Also save the node where we saw it.
+                        results[param_name] = param_value
+                        first_seen_node[param_name] = node_name
+
+        return results
```
mi-dai, line 384: General thought: do we want to give warnings/errors if the node/parameter names don't match any in the dictionary?

src/tdastro/graph_state.py
```diff
@@ -321,47 +321,67 @@ def extract_single_sample(self, sample_num):
...
-            use_nodes = None
-        elif isinstance(nodes, str):
-            use_nodes = set([nodes])
-        else:
-            use_nodes = set(nodes)
+        # If we are looking up a single parameter, but it into a list.
+        if isinstance(params, str):
+            params = [params]
+
+        # Go through all the parameters. If a parameters fill name is provided,
```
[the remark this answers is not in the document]
jeremykubica: Fixed

src/tdastro/graph_state.py
```diff
@@ -321,47 +321,67 @@ def extract_single_sample(self, sample_num):
...
-        elif isinstance(params, str):
-            use_params = set([params])
-        else:
-            use_params = set(params)
+        if len(single_params) == 0:
+            # Nothing else to do.
+            return results
 
-        values = {}
+        # Traverse the trnested dictionaries looking for cases where the parameter names match.
```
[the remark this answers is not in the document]
jeremykubica: Fixed

src/tdastro/graph_state.py
jeremykubica: Added

mi-dai (approved):

PR #173, opened 2024-10-16
Helper function for auto-applying colmap

jeremykubica:
The user should not need to know about the colmap in `OpSim`. This provides a quick check that will auto-apply the colmap if it is present and has the correct mapping.


hombit (approved):
Nice

PR #175, opened 2024-10-24
Sample redshift distribution calculated from rate model

mi-dai:


docs/notebooks/test_snia.ipynb
```diff
@@ -96,7 +96,8 @@
     "res, passbands = run_snia_end2end(\n",
```
hombit, line 96: <pre class="ql-syntax" spellcheck="false">Line #5.        solid_angle=0.002,
</pre><p>Is it LSST's field of view in steradians? Could you please put a comment here.</p><p> </p><p>Also we use field-of-view radius when matching positions with opsim, should we use introduce these numbers as constants in the code?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tdastro/pull/175/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

hombit, line 96: <pre class="ql-syntax" spellcheck="false">Line #5.    plt.plot(z2, nsn * 0.0005)
</pre><p>What does this 0.0005 mean?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tdastro/pull/175/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

mi-dai: <p>Random solid angle to get a reasonable number for testing. That sounds good to me to put field of view for LSST as constant.</p>

mi-dai: <p>Random number again to look at the distributions in the same scale (me being lazy not renormalizing any of those). I need to change this to use the actual volumetric_rate*dV function instead of the integrated one, but not for this PR. With larger number of bins the function is close enough for testing purpose right now.</p>

hombit (approved):
All sound good, feel free to merge

PR #178, opened 2024-10-25
Refactor test_snia.py

jeremykubica:
Refactor test_snia.py to break up the code and make it easier to debug/profile. The only behavioral change is making the comparison with a direct sncosmo computation optional. This speeds things up a bit during simulation.

A few stylistic cleanups:
- Adopts the new parameter access methods, so we do not have to go through the node's `get_param` function.
- Samples all runs at the beginning instead of in the core loop. This is more efficient for large number of samples.

### DIFF
--- a/src/tdastro/example_runs/simulate_snia.py
+++ b/src/tdastro/example_runs/simulate_snia.py
@@ -177,7 +177,7 @@ def draw_single_random_sn(
     # match band names in passbands object.
     times = opsim["time"][obs_index].to_numpy()
     if len(times) == 0:
-        raise ValueError(f"No overlap time in opsim for (ra,dec)=({ra:.2f},{dec:.2f})")
+        logger.warning(f"No overlap time in opsim for (ra,dec)=({ra:.2f},{dec:.2f})")
     res["times"] = times
 
     filters = opsim["filter"][obs_index].to_numpy(str)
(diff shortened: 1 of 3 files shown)


src/tdastro/example_runs/simulate_snia.py
```diff
@@ -16,48 +19,173 @@
...
     times = opsim["time"][obs_index].to_numpy()
-    filters = opsim["filter"][obs_index].to_numpy(str)
-    # Change to match band names in passbands object
-    filters = np.char.add("LSST_", filters)
-
     if len(times) == 0:
-        print(f"No overlap time in opsim for (ra,dec)=({ra:.2f},{dec:.2f})")
-        return
-
+        raise ValueError(f"No overlap time in opsim for (ra,dec)=({ra:.2f},{dec:.2f})")
```
mi-dai, line 180: I don't think we want error here because we want to keep the loop running.

jeremykubica: I made it a warning. Let me know if you'd like me to move it back to a print statement or debug logging or something else.

mi-dai (approved):

src/tdastro/example_runs/simulate_snia.py
mi-dai: Warning should work for now. Thanks!

PR #179, opened 2024-10-28
Speed up unit conversions

jeremykubica:
Do a single matrix multiply instead of carrying the units through the full computation.

### DIFF
--- a/src/tdastro/astro_utils/unit_utils.py
+++ b/src/tdastro/astro_utils/unit_utils.py
@@ -48,8 +48,8 @@ def flam_to_fnu(flux_flam, wavelengths, *, wave_unit, flam_unit, fnu_unit):
     flux_fnu : `list` or `np.array`
         The flux values in fnu units.
     """
-    flux_flam = np.array(flux_flam)
-    wavelengths = np.array(wavelengths)
+    flux_flam = np.asarray(flux_flam)
+    wavelengths = np.asarray(wavelengths)
 
     # Check if we need to reshape wavelengths to match the number
     # of rows in flux_flam.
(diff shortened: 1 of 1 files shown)


src/tdastro/astro_utils/unit_utils.py
```diff
@@ -27,8 +48,8 @@ def flam_to_fnu(flux_flam, wavelengths, *, wave_unit, flam_unit, fnu_unit):
     flux_fnu : `list` or `np.array`
         The flux values in fnu units.
     """
-    flux_flam = np.array(flux_flam) * flam_unit
-    wavelengths = np.array(wavelengths) * wave_unit
+    flux_flam = np.array(flux_flam)
+    wavelengths = np.array(wavelengths)
```
hombit, line 52: Since we touch this code, let's switch to `asarray` - it doesn't copy data by default
```suggestion
    flux_flam = np.asarray(flux_flam)
    wavelengths = np.asarray(wavelengths)
```

src/tdastro/astro_utils/unit_utils.py
```diff
@@ -73,8 +95,8 @@ def fnu_to_flam(flux_fnu, wavelengths, *, wave_unit, flam_unit, fnu_unit):
     flux_flam : `list` or `np.array`
         The flux values in flam units.
     """
-    flux_fnu = np.array(flux_fnu) * fnu_unit
-    wavelengths = np.array(wavelengths) * wave_unit
+    flux_fnu = np.array(flux_fnu)
+    wavelengths = np.array(wavelengths)
```
hombit, line 99:
```suggestion
    flux_fnu = np.asarray(flux_fnu)
    wavelengths = np.asarray(wavelengths)
```

hombit (approved):
Looks good, thank you. How much faster is this?

src/tdastro/astro_utils/unit_utils.py
jeremykubica: Good call.  I made these changes.

PR #180, opened 2024-10-29
[DRAFT PR] Possible dust map implementation

jeremykubica:
Draft PR to ask if this is the direction we want to go.


src/tdastro/astro_utils/dust_map.py
```diff
@@ -0,0 +1,141 @@
...
+            Default = 1.0
+
+        Returns
+        -------
+        flux_density : `numpy.ndarray`
+            The results (in nJy).
+        """
+        # Get the extinction value at the object's location.
+        coord = SkyCoord(ra, dec, dist, frame="icrs", unit="deg")
+        ebv = self.dust_map.query(coord)
```
hombit, line 136: The problem with this approach is that different maps return different values - it is not always E(B-V). An example:
https://dustmaps.readthedocs.io/en/latest/modules.html#module-dustmaps.bayestar

src/tdastro/astro_utils/dust_map.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Do we need to convert ebv by a factor from this table:
+        # https://iopscience.iop.org/article/10.1088/0004-637X/737/2/103#apj398709t6
+
+        return flux_density * self.extinction_model.extinguish(wavelengths * u.angstrom, Ebv=ebv)
```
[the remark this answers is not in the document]
jeremykubica: Yes. All models in the dust_extinction package use `extinguish`

src/tdastro/astro_utils/dust_map.py
```diff
@@ -109,33 +116,53 @@ def load_extinction_model(name, **kwargs):
...
+        provide a ebv value directly or (RA, dec) and distance information that can
+        be used in the dustmap query.
 
         Parameters
         ----------
-        flux_density : `numpy.ndarray`
+        flux_density : numpy.ndarray
             An array of flux density values (in nJy).
-        wavelengths : `numpy.ndarray`, optional
+        wavelengths : numpy.ndarray, optional
```
mi-dai, line 128: Why is wavelengths optional here?

src/tdastro/astro_utils/dust_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Add the correction function.
+        def _sfd_scale_ebv(input, **kwargs):
+            """Scale the result of the SFD query."""
+
+        self.ebv_func = _sfd_scale_ebv
+
+    @staticmethod
+    def load_extinction_model(name, **kwargs):
+        """Load the extinction model.
```
mi-dai, line 98: I suggest add source of the extinction model for users who are not familiar with the package.
"Load the extinction model from the dust_extinction package (https://github.com/karllark/dust_extinction)"

PR #181, opened 2024-10-30
Create a new node type that wraps pzflow and create an introductory notebook

jeremykubica:
Create a new node type that allows us to consistently sample from `pzflow`. This node will determine its parameters based on the information the flow computes, allowing them to be used for other parameter setters.

The demo notebook includes the basics of setting up nodes and sampling. It uses the new pzflow node as an example.


mi-dai (approved):
Looks good to me! Sorry again for the delay. 
One comment for the notebook - I think it could work better with a real example of astronomical source instead of the StaticSource, but we can do this later.

PR #182, opened 2024-10-31
Improve error checking for node labels

jeremykubica:
The node strings must be unique in order to store the data internally in a `GraphState`. This PR changes the logic to check that each node seen during sampling has a unique string.

Also adds a few very small performance optimizations (e.g. `np.array` -> `np.asarray`).


OliviaLynn (approved):
Looks good!

PR #185, opened 2024-11-04
Create a BasicMathNode that can evaluate expressions given as a string

jeremykubica:
Creates a new `BasicMathNode` that will evaluate basic math functions given as a string. Performs a bunch of checks to prevent arbitrary code execution. Also translates math functions into the correct backend (`math`, `numpy`, or `jax`).

This will allow us to write functions nodes (for basic math) as:
```
x0_node = BasicMathNode(
    "pow(10.0, -0.4 * (distmod - alpha * x1 + beta * c + m_abs - 10.635))",
    distmod = host.distmod,
    alpha=...
    ...
)
```

### DIFF
--- a/src/tdastro/math_nodes/basic_math_node.py
+++ b/src/tdastro/math_nodes/basic_math_node.py
@@ -1,10 +1,17 @@
-"""Nodes that perform basic math operations."""
+"""Nodes that perform basic math operations that can be specified as strings.
+
+The goal of this library is to save users from needing to create a bunch of
+small FunctionNodes to perform basic math.
+"""
 
 import ast
-import math
 
-import jax.numpy as jnp
-import numpy as np
+# Disable unused import because we need all of these imported
+# so they can be used during evaluation of the node.
+import math  # noqa: F401
+
+import jax.numpy as jnp  # noqa: F401
+import numpy as np  # noqa: F401
 
 from tdastro.base_models import FunctionNode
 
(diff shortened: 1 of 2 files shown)


src/tdastro/math_nodes/basic_math_node.py
```diff
@@ -0,0 +1,185 @@
...
+# so they can be used during evaluation of the node.
+import math  # noqa: F401
+
+import jax.numpy as jnp  # noqa: F401
+import numpy as np  # noqa: F401
+
+from tdastro.base_models import FunctionNode
+
+
+class BasicMathNode(FunctionNode):
```
mtauraso, line 19: Totally irrelevant to this PR, but as a grad student I've definitely wished for this sort of conversion as a callable library, especially if C stdlib was one of the "basic math" languages that could be converted.

hombit (approved):
Some general usability issues, most of them are not related to this PR and can be discussed separately:

1. `BasicMathNode` doesn't have a string representation
2. `GraphState` doesn't have a string representation and iteration over it fails, e.g. `list(BasicMathNode("a", a=1, node_label="node").sample_parameters())`
3. `BasicMathNode("a", a=1).sample_parameters()` cannot be used with `[]` syntax because `node_label is None`. However I still could use `extract_parameters`
4. Math exceptions like `1/0` or `log(-1)` are hard to track
5. It is not clear how I can a) jit-compile with `jax` and b) use a random number node.

jeremykubica:
Thanks @hombit, I filed some issues for some of those:
1) I need to think about how to provide a string representation, because I currently use `__str__` for some of the node labeling.
2) Added issue #186 
3) Added issue #187
4) Agreed. I don't have a great solution here, but I can try adding better debugging (#188)
5a) I don't think jit compiling is possible with this node, so we would need to be careful. 
5b) You can use a random variable as input the same way as within any other node.

PR #190, opened 2024-11-07
Add some basic logging for errors in math node

### LINKED ISSUE #188: Add better exception handling for math errors
Extend the `BasicMathNode` to provide better error messages for math exceptions such as divide by zero or negative log.

jeremykubica:
closes #188 

Add logging to record the expression being evaluated and the parameters used when an exception is encountered in a `BasicMathNode`.

### DIFF
--- a/src/tdastro/math_nodes/basic_math_node.py
+++ b/src/tdastro/math_nodes/basic_math_node.py
@@ -133,11 +133,8 @@ class BasicMathNode(FunctionNode):
             except Exception as problem:
                 # Provide more detailed logging, including the expression and parameters
                 # used, when we encounter a math error like divide by zero.
-                logger.error(
-                    f"{type(problem)} encountered during operation: {self.expression}\n"
-                    f"with arguments={kwargs}"
-                )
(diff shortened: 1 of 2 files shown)


src/tdastro/math_nodes/basic_math_node.py
```diff
@@ -125,7 +128,16 @@ def __init__(self, expression, backend="numpy", node_label=None, **kwargs):
...
+            try:
+                return eval(self.expression, globals(), kwargs)
+            except Exception as problem:
+                # Provide more detailed logging, including the expression and parameters
+                # used, when we encounter a math error like divide by zero.
+                logger.error(
+                    f"{type(problem)} encountered during operation: {self.expression}\n"
+                    f"with arguments={kwargs}"
+                )
+                raise problem
```
hombit, line 140: Could we add some context to an exception and raise with a better message and/or traceback?

hombit (approved):
Great idea, thank you! I wonder if we could also put this information into the traceback/exception message.

PR #191, opened 2024-11-21
Initial Implementation of a JAX Salt2 model

jeremykubica:
Create an initial implementation of the SALT2 model using JAX. This will run slower on CPU (~4x on my laptop), but should be able to take advantage of both auto-differentiation and GPU/TPU. Further optimizations are likely possible, but this gets an initial base implementation checked in.


hombit (approved):
1. When I specify wrong model directory for `SALT2JaxModel` I'm getting `InconsistentTableError` instead of `FileNotFoundError` raised by `SALT2Source`
2. When trying to run `.evaluate / .get_band_fluxes` with array parameters for x0, x1, c I've got broadcasting error. Actually it looks like we have similar problems across the code, for example for passbands. So it could go to a new issue.

jeremykubica:
I fixed the first bug (adding a FileNotFoundError). The second one is a known limitation. I have a mental todo list items to better support batch evaluations. I will add a github issue for that.

PR #194, opened 2024-11-22
Expand the introductory notebook

jeremykubica:
Expand the introductory notebook to include more discussion on FunctionNodes including the BasicMathNode.

In the process fix a bug with the basic math node where it would give an error if used in parameter chaining (by virtue of being callable itself).


docs/notebooks/introduction_demo.ipynb
```diff
@@ -115,7 +106,9 @@
...
+    "The sampling process create **vectors** of samples for each parameter such that the `i`-th value of each parameter is from the same sampling run. So in the output above, sample 0 consists of all the parameter values for that sample (everything at index=0), sample 1 consists of all parameter values for that sample (everything at index=1), and so forth. This is critically important once we start dealing with parameters are not independent. We might want to choose host galaxy's mass and redshift from a joint distribution. Alternatively, as we will see below, we will often want to compute one parameter as a mathematical transform of another or sample one parameter based on the value of another. In all of these cases it is important that we can access the parameters that were generated together and that these parameters stay consistent.\n",
```
OliviaLynn, line 109:
```suggestion
    "The sampling process creates **vectors** of samples for each parameter such that the `i`-th value of each parameter is from the same sampling run. So in the output above, sample 0 consists of all the parameter values for that sample (everything at index=0), sample 1 consists of all parameter values for that sample (everything at index=1), and so forth. This is critically important once we start dealing with parameters that are not independent. We might want to choose the host galaxy's mass and redshift from a joint distribution. Alternatively, as we will see below, we will often want to compute one parameter as a mathematical transform of another or sample one parameter based on the value of another. In all of these cases it is important that we can access the parameters that were generated together and that these parameters stay consistent.\n",
```

docs/notebooks/introduction_demo.ipynb
```diff
@@ -132,7 +125,105 @@
...
+    "redshifts = np.array([0.1, 0.2, 0.3])\n",
+    "distmods = cosmo_obj.distmod(redshifts).value\n",
+    "print(distmods)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Importantly, you can use a `FunctionNode` that takes input parameter(s) and produces output parameter(s) during the generation `FunctionNode` is a subclass of `ParameterizedNode` that wraps the functionality of collecting the inputs, running the computations, and storing the output. The user simply needs to provide the function node with the function it will use and the parameters. For example, TDAstro has a `DistModFromRedshift` class that wraps the previous operation.\n",
```
OliviaLynn, line 158:
```suggestion
    "Importantly, you can use a `FunctionNode` that takes input parameter(s) and produces output parameter(s) during the generation. `FunctionNode` is a subclass of `ParameterizedNode` that wraps the functionality of collecting the inputs, running the computations, and storing the output. The user simply needs to provide the function node with the function it will use and the parameters. For example, TDAstro has a `DistModFromRedshift` class that wraps the previous operation.\n",
```

OliviaLynn: Hard to see, but adding a period after "generation"

docs/notebooks/introduction_demo.ipynb
```diff
@@ -132,7 +125,105 @@
...
+    "distmod_obj = DistModFromRedshift(\n",
+    "    H0=73.0, Omega_m=0.3, redshift=NumpyRandomFunc(\"uniform\", low=0.1, high=0.5)\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "or more concretely, we can create out own `FunctionNode` that computes y = m * x + b."
```
OliviaLynn, line 180:
```suggestion
    "or more concretely, we can create our own `FunctionNode` that computes y = m * x + b."
```

OliviaLynn (approved):
Looks good!

PR #202, opened 2024-12-20
Break out plotting functions

jeremykubica:
Break the plotting functions out from the passbands demo notebook so that we can use them in other notebooks or for analysis. This also fixes a bug in the notebook where the wavelength range of the spline model did not overlap the wavelength range of the passbands.


OliviaLynn (approved):
Looks good!

PR #206, opened 2025-01-02
Create a tutorial notebook for Opsim

jeremykubica:
Create a tutorial notebook for the `OpSim` class and add some helper function that would be useful for the snia code, including:
- Getting time bounds from OpSim
- Filtering OpSim on a column
- Creating random OpSim


src/tdastro/astro_utils/opsim.py
```diff
@@ -301,6 +301,49 @@ def write_opsim_table(self, filename, tablename="observations", overwrite=False)
...
+        Returns
+        -------
+        t_min, t_max : float, float
+            The min and max times for all observations in the OpSim.
+        """
+        t_min = self.table[self.colmap["time"]].min()
+        t_max = self.table[self.colmap["time"]].max()
+        return t_min, t_max
+
+    def filter_rows(self, rows):
```
hombit, line 316: I'd prefer to return a new OpSim object here.

hombit (approved):
Thank you! I have a couple of minor comments about the code; the notebook looks great.

src/tdastro/astro_utils/opsim.py
```diff
@@ -403,6 +446,58 @@ def bandflux_error_point_source(self, bandflux, index):
         )
 
 
+def create_random_opsim(num_obs, seed=None):
```
hombit, line 449: This is very close to what we have at https://github.com/lincc-frameworks/tdastro/tree/main/tests/tdastro/testing_utils. Should we move things around so all relevant functions are in one place? Either here or in that directory.

jeremykubica: We should eventually combine them. Let's discuss which of the locations makes the most sense and do that in a later PR.

src/tdastro/astro_utils/opsim.py
jeremykubica: Done.

PR #207, opened 2025-01-02
Extend random nodes for testing and simulation

jeremykubica:
This PR adds a bunch of new random generation functions (for testing and simulation) and also makes some small clean ups to how random generating function nodes are defined in the codebase. 

Changes include:
- Merging the `generate()` function defined in almost every random node with the existing `resample_and_compute()` function defined in base_models.py. The new function is called `generate()` and supports both generating samples (as used for the random nodes before) and computing gradients (as used in `resample_and_compute()`).  This removes a bunch of duplicate code and makes creating new random generating nodes a little easier.
- Create `UniformRADEC` node that generates uniform (RA, dec) samples over entire sphere. This can be used for testing or simulation.
- Create `OpSimRADECSampler` node that generates uniform (RA, dec, time) samples from a given OpSim data base. This can be used for testing or simulation.
- Renames `GivenSampler` to `GivenValueList` since the node was iterating over a list (not sampling from it).
- Creates a `GivenValueSampler` that does sample from a list of values.
- Updates the `TableSampler` node to either iterate over the values in order or sample randomly from them. The random sampling forms the core of the `OpSimRADECSampler` node's behavior.

### DIFF
--- a/src/tdastro/math_nodes/given_sampler.py
+++ b/src/tdastro/math_nodes/given_sampler.py
@@ -10,8 +10,8 @@ from astropy.table import Table
 from tdastro.base_models import FunctionNode
 
 
-class GivenSampler(FunctionNode):
-    """A FunctionNode that returns given results.
+class GivenValueList(FunctionNode):
+    """A FunctionNode that returns given results for a single parameter.
 
     Attributes
     ----------
--- a/src/tdastro/math_nodes/given_sampler.py
+++ b/src/tdastro/math_nodes/given_sampler.py
@@ -8,6 +8,7 @@ import pandas as pd
 from astropy.table import Table
 
 from tdastro.base_models import FunctionNode
+from tdastro.math_nodes.np_random import NumpyRandomFunc
 
 
 class GivenValueList(FunctionNode):
@@ -22,8 +23,11 @@ class GivenValueList(FunctionNode):
     """
 
     def __init__(self, values, **kwargs):
-        self.values = np.array(values)
+        self.values = np.asarray(values)
+        if len(values) == 0:
+            raise ValueError("No values provided for GivenValueList")
         self.next_ind = 0
+
         super().__init__(self._non_func, **kwargs)
 
     def _non_func(self):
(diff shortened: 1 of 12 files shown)


src/tdastro/math_nodes/given_sampler.py
```diff
@@ -73,42 +77,108 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+            functions can easily access the results.
+        """
+        rng = rng_info if rng_info is not None else self._rng
+
+        if graph_state.num_samples == 1:
+            inds = rng.integers(0, self._num_values)
+        else:
+            inds = rng.integers(0, self._num_values, size=graph_state.num_samples)
+
+        return self.values[inds]
```
hombit, line 127: I'm not sure if it is a better match here, but `rng.choice(values, size=...)` would do the same job as `self.value[rng.integers]`.

src/tdastro/math_nodes/given_sampler.py
```diff
@@ -73,42 +77,108 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+        self._num_values = len(self.data)
+        if self._num_values == 0:
+            raise ValueError("No data provided to TableSampler.")
 
-    def _non_func(self):
-        """This function does nothing. Everything happens in the overloaded compute()."""
-        pass
+        # Add each of the flow's data columns as an output parameter.
+        self.columns = [x for x in self.data.colnames]
+        super().__init__("uniform", outputs=self.columns, **kwargs)
```
hombit, line 178: I don't think we need a new list here 
```suggestion
        super().__init__("uniform", outputs=self.data.colnames, **kwargs)
```

src/tdastro/math_nodes/ra_dec_sampler.py
```diff
@@ -0,0 +1,142 @@
...
+        -------
+        results : any
+            The result of the computation. This return value is provided so that testing
+            functions can easily access the results.
+        """
+        rng = rng_info if rng_info is not None else self._rng
+
+        # Generate the random (RA, dec) lists.
+        ra = rng.uniform(0.0, 2.0 * np.pi, size=graph_state.num_samples)
+        dec = np.arccos(2.0 * rng.uniform(0.0, 1.0, size=graph_state.num_samples) - 1.0) - (np.pi / 2.0)
```
hombit, line 52: A bit more readable
```suggestion
        dec = np.arcsin(rng.uniform(-1.0, 1.0, size=graph_state.num_samples))
```

src/tdastro/math_nodes/given_sampler.py
jeremykubica: Agreed. I'm not sure if one is better than the other, so I will leave it as-is for now. We can always change it in the future if needed.

PR #208, opened 2025-01-02
Expand documentation of the inherited parameterized values

jeremykubica:
There is probably a better way to do this using meta-programming and inheriting part of the docstrings. But for now, we just list out the parameterized values that different `PhysicalModel` nodes have. This will help new users know what can be set.

Also add two helper functions:
- `list_params()` is for interactive development and debugging if a user wants to check what parameters are there (any node, not just a Physical Model).
- `has_valid_param()` performs a contains and a is not None check


hombit (approved):
It is very useful - thank you!

PR #209, opened 2025-01-02
Move t0 into PhysicalModel

jeremykubica:
Originally we decided not to include a `t0` parameter in the `PhysicalModel` class, because it won't be used by all models. However we currently fail if we apply redshift without a valid `t0`. This PR includes `t0` in the base `PhysicalModel` and sets it to 0.0 (no effect) by default.

We also use the occurrence of `t0` to allow filtering by time to match the time pruning done in @mi-dai 's notebook in #205. By default the `PhysicalModel` uses all times. But subclasses, such as sncosmo models, can add their own restrictions.

### DIFF
--- a/src/tdastro/sources/periodic_variable_star.py
+++ b/src/tdastro/sources/periodic_variable_star.py
@@ -31,7 +31,7 @@ class PeriodicVariableStar(PeriodicSource, ABC):
 
     def __init__(self, period, **kwargs):
         super().__init__(period, **kwargs)
-        if "distance" not in self.setters:
+        if not self.has_valid_param("distance"):
             raise ValueError("Distance parameter is required for PeriodicVariableStar")
 
     def _evaluate_phases(self, phases, wavelengths, graph_state, **kwargs):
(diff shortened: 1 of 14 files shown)


src/tdastro/sources/periodic_variable_star.py
```diff
@@ -29,9 +29,9 @@ class PeriodicVariableStar(PeriodicSource, ABC):
         The distance to the source, in pc.
     """
 
-    def __init__(self, period, t0, **kwargs):
-        super().__init__(period, t0, **kwargs)
-        if not self.has_valid_param("distance"):
+    def __init__(self, period, **kwargs):
+        super().__init__(period, **kwargs)
+        if "distance" not in self.setters:
```
hombit, line 34: Not an issue, but a question: why this change?

hombit (approved):

src/tdastro/sources/periodic_variable_star.py
jeremykubica: That is actually a merge artifact from working on this at the same time I was working on the `has_valid_param` I sent you earlier today. I am changing it back to `has_valid_param` since that also does the `None` check.

PR #210, opened 2025-01-02
Add a helper function for assembling a snia graph

jeremykubica:
Instead of having users always manually recreate all the nodes needed for each graph, we can prebuild sections of the graph. `snia_x0_x1_from_host()` creates and links the nodes used by snia models.

### DIFF
--- a/docs/notebooks/opsim_notebook.ipynb
+++ b/docs/notebooks/opsim_notebook.ipynb
@@ -292,6 +292,34 @@
     "%%timeit\n",
     "_ = opsim_data2.range_search(query_ra, query_dec, 0.5)"
    ]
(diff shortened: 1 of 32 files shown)


mi-dai (commented):
This looks good to me. My concern is now we lose the flexibility to define `x1_func` in other ways since it's already pre-defined in this `snia_x0_x1_from_host` function.

jeremykubica:
It is a trade off. The question for you is whether it is worth having this as a default implementation or whether the user should specify it themselves in the simulation (as in your demo notebook). I think either way will be fine.

jeremykubica:
Sounds good. Closing this PR.

PR #211, opened 2025-01-02
Provide ability to use a default search radius for the OpSim

jeremykubica:
Allow a default search radius so we can pass all the information needed for spatial matching together in a single `OpSim` object.


hombit (approved):
Thank you

PR #212, opened 2025-01-03
Add helper functions for the passband

jeremykubica:
Restructures the passband group based on how it is being used in the initial notebooks. Key changes:
- Cleanup: `Passband` no longer stores table URL or file path. Instead these are passed to the relevant loader functions.
- Cleanup: All passband data is stored in a top level data directory instead of in the src directory. This new directory is added to gitignore so the passbands are not committed to git.
- Ease of use: Added ability to lookup passband by filter name from the `PassbandGroup`.
- Ease of use: Added ability to load only a subset of the filters.
- Introduction: Extended the introduction notebook.

### DIFF
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -6,12 +6,25 @@ from pathlib import Path
 from typing import Literal, Optional, Union
 from urllib.error import HTTPError, URLError
 
+import matplotlib.pyplot as plt
 import numpy as np
 import scipy.integrate
 
 import tdastro
 
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
+
+
+# Set default colors for plotting to match:
+# https://community.lsst.org/t/lsst-filter-profiles/1463
+_lsst_filter_plot_colors = {
+    "g": "blue",
+    "i": "yellow",
+    "r": "green",
+    "u": "purple",
+    "y": "red",
+    "z": "orange",
+}
 
 
 class PassbandGroup:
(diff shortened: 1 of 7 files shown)


src/tdastro/astro_utils/passbands.py
```diff
@@ -643,3 +880,68 @@ def fluxes_to_bandflux(
...
+        )
+        if plot_loaded:
+            ax.plot(
+                self._loaded_table[:, 0],  # X values are the wavelength
+                self._loaded_table[:, 1],  # Y values are the transmission values.
+                color=color,
+                linestyle="--",
+            )
+
+        ax.set_xlabel("Wavelength (A)")
```
hombit, line 922: It works even with no TeX installation:
```suggestion
        ax.set_xlabel(r"Wavelength, $\AA$")
```

src/tdastro/astro_utils/passbands.py
```diff
@@ -643,3 +880,68 @@ def fluxes_to_bandflux(
         integrand = flux_density_matrix * self.processed_transmission_table[:, 1]
         bandfluxes = scipy.integrate.trapezoid(integrand, x=self.waves, axis=w_axis)
         return bandfluxes
+
+    def plot(self, ax=None, figure=None, color=None, plot_loaded=False):
```
hombit, line 884: I'd also add `ax.set_ylim(0, None)`

src/tdastro/astro_utils/passbands.py
```diff
@@ -6,15 +6,37 @@
...
+
+# Set default colors for plotting to match:
+# https://community.lsst.org/t/lsst-filter-profiles/1463
+_lsst_filter_plot_colors = {
+    "g": "blue",
+    "i": "yellow",
+    "r": "green",
+    "u": "purple",
+    "y": "red",
+    "z": "orange",
```
hombit, line 27: Wavelength order:
```suggestion
    "u": "purple",
    "g": "blue",
    "r": "green",
    "i": "yellow",
    "z": "orange",
    "y": "red",
```

hombit (commented):
LGTM. I'm not pushing "Approve" to let @OliviaLynn make it green

PR #213, opened 2025-01-03
Add bounds checking for sncosmo models

jeremykubica:
Instead of crashing when given wavelengths out of bounds, return zero flux in those wavelengths.

### DIFF
--- a/tests/tdastro/sources/test_sncosmo_models.py
+++ b/tests/tdastro/sources/test_sncosmo_models.py
@@ -73,7 +73,7 @@ def test_sncomso_models_hsiao_t0() -> None:
 
 
 def test_sncomso_models_bounds() -> None:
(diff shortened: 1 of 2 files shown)


tests/tdastro/sources/test_sncosmo_models.py
```diff
@@ -72,6 +72,28 @@ def test_sncomso_models_hsiao_t0() -> None:
     assert np.array_equal(mask, expected_mask)
 
 
+def test_sncomso_models_bounds() -> None:
+    """Test that wwe do not crash is we give wavelengths outside the model bounds."""
```
mi-dai, line 76:
```suggestion
    """Test that we do not crash if we give wavelengths outside the model bounds."""
```

mi-dai (approved):
This is good for now. We should allow users to fill the out-of-bound values the way they want in the future.

PR #218, opened 2025-01-08
[Draft] Extension of Mi's Notebook

jeremykubica:
This PR converts the notebook in #205 to use the new functions.

Co-authored-by: @mi-dai


OliviaLynn (commented):
Took at look through this and it looks really good to me—nothing jumps out to me (of course, as a non-scientist) as something that can be further simplified for the user (but interested to hear otherwise!)

docs/notebooks/plasticc_snia.ipynb
```diff
@@ -0,0 +1,234 @@
+{
```
mi-dai, line 1: <p>I think this is not optional because we currently need the opsim filters and the passband filters match each other otherwise it will error</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tdastro/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

docs/notebooks/plasticc_snia.ipynb
```diff
@@ -0,0 +1,234 @@
...
+    "passband_group = PassbandGroup(preset=\"LSST\", filters_to_load=[\"g\", \"r\", \"i\", \"z\"])\n",
+    "print(f\"Loaded Passbands: {passband_group}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ab9d7cf5",
+   "metadata": {},
+   "source": [
+    "We don't care about any of the observations in the OpSim that are not in one of the filters of interest. We can use a combination of `PassbandGroup`'s `mask_by_filter()` to determine which observations match and `OpSim`'s `filter_rows()` to remove those rows that do not match. This is an optional step that will improve the speed at which the algorithm runs by removing unused data."
```
mi-dai, line 73: I think this is not optional because we currently need the opsim filters and the passband filters match each other otherwise it will error

mi-dai (commented):
Looks really good to me. One thing that may help in the `plot_lightcurves` function is to take the nested df as input instead of the flux, times, filters columns, but not an urgent need, since we may also want other plotting functions later.

docs/notebooks/plasticc_snia.ipynb
jeremykubica: Do you mean that it should throw an error instead? Or something else?

mi-dai: I meant the description was not accurate. Users will get an error if they don't do the filter, so it's not "optional"

jeremykubica: Good call. I removed that sentence.

jeremykubica:
Changes added in #205

PR #220, opened 2025-01-21
Update copier version and add sphinx docs.

delucchi-cmu:
Updates to add basic sphinx docs. Moves some notebooks into "pre_executed" subdirectory, so they are not executed in CI or on RTD, as they're slow and require external resources.


docs/notebooks/introduction_demo.ipynb
```diff
@@ -376,7 +376,7 @@
     "from pzflow import Flow\n",
     "from tdastro import _TDASTRO_TEST_DATA_DIR\n",
     "\n",
-    "pzflow = Flow(file=_TDASTRO_TEST_DATA_DIR / \"test_flow.pkl\")\n",
+    "pzflow = Flow(file=\"../../tests/tdastro/data/test_flow.pkl\")\n",
```
jeremykubica, line 379: Why change the path? Is this needed for auto-notebook rendering?

docs/notebooks/pre_executed/test_snia.ipynb
```diff
@@ -1,5 +1,13 @@
 {
  "cells": [
+  {
```
jeremykubica, line 3: We can remove this notebook altogether. It has been replaced by the plastic_snia.ipynb notebook.

jeremykubica (commented):
Small comments / questions

docs/notebooks/introduction_demo.ipynb
delucchi-cmu: Yes - the path doesn't resolve the same way in CI and RTD as it does in unit tests, so the file isn't found.

jeremykubica (approved):

PR #224, opened 2025-01-24
Add the ability to sample from an empirically defined distribution.

jeremykubica:
Add FunctionNode the performs a similar functionality to SNANA's `DistributionSampler`, but uses different interpolation to avoid values ever going below 0.0. This version may be less accurate in some cases.


hombit (approved):
Looks good. I know that there are different ways to interpolate CDF (corresponding to different types of quantiles). I'd appreciate if @aimalz could also review it

jeremykubica:
Chatted with Alex in person. We are good to merge this in, but should look at bringing in QP long term.

PR #226, opened 2025-01-24
Add a white noise effect into the base PhysicalModel

jeremykubica:
Add a white noise parameter to `PhysicalModel` so that any model can have white noise applied (it is off be default).

This PR also reorganizes the basic models by moving `StaticModel` and `StepModel` into `basic_models.py`. Also creates a `SinWaveModel` in `basic_models.py`.


src/tdastro/sources/physical_model.py
```diff
@@ -199,9 +224,20 @@ def evaluate(self, times, wavelengths, graph_state=None, given_args=None, rng_in
                     times, wavelengths, params["redshift"], params["t0"]
                 )
 
-            # Compute the flux density for both the current object and add in anything
-            # behind it, such as a host galaxy.
+            # Compute the flux density for both the current object, add and white noise,
+            # and add in in anything behind it, such as a host galaxy.
```
OliviaLynn, line 228:
```suggestion
            # Compute the flux density for the current object, then add white noise,
            # and then add in anything behind the object, such as a host galaxy.
```

tests/tdastro/sources/test_basic_sources.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # We can set a value we have already added.
+    model.set_parameter("brightness", 5.0)
+    values = model.evaluate(times, wavelengths)
+    assert values.shape == (6, 3)
+    assert np.all(values == 5.0)
+
+
+def test_static_source_pytree():
+    """Test tthat the PyTree only contains brightness."""
```
OliviaLynn, line 48:
```suggestion
    """Test that the PyTree only contains brightness."""
```

OliviaLynn (approved):
Looks good!

PR #227, opened 2025-01-25
Fix flaky test

jeremykubica:
One of the tests is failing because it uses random numbers that occasionally result in bad choices. This PR fixes the flaky test by setting a seed for the random number generator.


OliviaLynn (approved):
Nice catch!

PR #245, opened 2025-02-17
Fix the redshift computation with multiple samples

jeremykubica:
The current redshift computation is broken for multiple samples because it does not maintain a constant observer frame copy of wavelengths or time. Instead both arrays are redshifted on each iteration. This fixes that problem by creating new `rest_time` and `rest_wave` copies to represent the redshifted values.

It also passes times (optionally) to the `EffectsModels` (which is how I found the problem).

### DIFF
--- a/tests/tdastro/sources/test_physical_models.py
+++ b/tests/tdastro/sources/test_physical_models.py
@@ -108,7 +108,7 @@ def test_physical_model_evaluate():
 
 
 def test_physical_model_evaluate_redshift():
-    """Test that is we apply redshift to a model we get different flux values."""
+    """Test that if we apply redshift to a model we get different flux values."""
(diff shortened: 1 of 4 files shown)


tests/tdastro/sources/test_physical_models.py
```diff
@@ -95,6 +107,21 @@ def test_physical_model_evaluate():
     assert np.all(flux[2, :, :] == 30.0)
 
 
+def test_physical_model_evaluate_redshift():
+    """Test that is we apply redshift to a model we get different flux values."""
```
OliviaLynn, line 111:
```suggestion
    """Test that if we apply redshift to a model we get different flux values."""
```

OliviaLynn (approved):
Looks good!

PR #246, opened 2025-02-17
Add an extinction effect

jeremykubica:
Create an effect the applies a dust extinction given an ebv. This will later be chained with an effect that computes the ebv from the dustmap. So effectively this performs part of what #180 is meant to do. But by breaking this into two stages, we can be a lot more flexible.

### DIFF
--- a/src/tdastro/effects/extinction.py
+++ b/src/tdastro/effects/extinction.py
@@ -26,7 +26,7 @@ class ExtinctionEffect(EffectModel):
         Any additional keyword arguments.
     """
 
-    def __init__(self, extinction_model, ebv=None, **kwargs):
+    def __init__(self, extinction_model="F99", ebv=None, **kwargs):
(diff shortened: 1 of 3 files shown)


src/tdastro/effects/extinction.py
```diff
@@ -0,0 +1,119 @@
...
+        The extinction object from the dust_extinction library or its name.
+        If a string is provided, the code will find a matching extinction
+        function in the dust_extinction package and use that.
+    ebv : parameter
+        The setter (function) for the extinction parameter E(B-V).
+    **kwargs : `dict`, optional
+        Any additional keyword arguments.
+    """
+
+    def __init__(self, extinction_model, ebv=None, **kwargs):
```
hombit, line 29: Let's set some default model? `F99`?

hombit (approved):
Looks great, thank you

PR #249, opened 2025-02-24
Add a dust map wrapper

jeremykubica:
Add a wrapper for the dustmap package (without directly depending on that package). Include a test (constant) dustmap and a notebook of how to use effects in general.

The PR also:
- Fixes a small bug that would cause an error when the name of the output of a function node was not "function_node_result".
- Moves the effects from a list passed in the constructor to an `add_effect` function. This allows easier setup for effects that depend on model parameters, such as RA and dec.

### DIFF
--- a/src/tdastro/astro_utils/dustmap.py
+++ b/src/tdastro/astro_utils/dustmap.py
@@ -82,7 +82,8 @@ class DustmapWrapper(DustEBV):
 
 
 class ConstantHemisphereDustMap(DustEBV):
-    """A DustMap with a constant value in each hemisphere.
+    """A DustMap with a constant value in each hemisphere for testing
+    and debugging purposes.
 
     Attributes
     ----------
--- a/src/tdastro/astro_utils/dustmap.py
+++ b/src/tdastro/astro_utils/dustmap.py
@@ -115,8 +115,7 @@ class ConstantHemisphereDustMap(DustEBV):
         if np.isscalar(ra):
             return self.north_ebv if dec >= 0 else self.south_ebv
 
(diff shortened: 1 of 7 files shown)


src/tdastro/astro_utils/dustmap.py
```diff
@@ -0,0 +1,142 @@
...
+        Returns
+        -------
+        ebv : float or np.array
+            The E(B-V) value or array of values.
+        """
+        if np.isscalar(ra):
+            return self.north_ebv if dec >= 0 else self.south_ebv
+
+        ebv = np.full_like(ra, self.north_ebv)
+        ebv[np.asarray(dec) < 0] = self.south_ebv
```
hombit, line 119: Just a stylistic suggestion, please feel free to ignore
```suggestion
        ebv = np.where(np.asarray(dec) < 0, self.south_ebv, self.north_ebv)
```

src/tdastro/astro_utils/dustmap.py
```diff
@@ -0,0 +1,142 @@
...
+        coord = SkyCoord(ra, dec, frame="icrs", unit="deg")
+        dustmap_value = self._dust_map.query(coord)
+
+        if self._ebv_func is not None:
+            return self._ebv_func(dustmap_value)
+        return dustmap_value
+
+
+class ConstantHemisphereDustMap(DustEBV):
+    """A DustMap with a constant value in each hemisphere.
```
hombit, line 85: Maybe say that it is for debugging purposes only?

hombit (approved):
Looks great, and `ConstantHemisphereDustMap` is fun

PR #250, opened 2025-02-25
added ZTFOpsim class for simulating ZTF

mi-dai:


src/tdastro/astro_utils/ztf_opsim.py
```diff
@@ -0,0 +1,304 @@
...
+                            + readnoise**2*nexposure*npix
+                            + darkcurrent*npix*exptime*nexposure)
+                        = 0
+    flux = 12.5 + 0.5*sqrt(625
+                            + 100( sky*npix*Gain
+                            + readnoise**2*nexposure*npix
+                            + darkcurrent*npix*exptime*nexposure) )
+    zp = 2.5log(flux) + maglim
+
+    Parameters
```
jeremykubica, line 68: Do you have the units for any of these parameters? The more we can add, the better the documentation will be.

src/tdastro/astro_utils/ztf_opsim.py
```diff
@@ -0,0 +1,304 @@
...
+    ----------
+    maglim : float or ndarray
+        Five-sigma magnitude limit.
+    sky : float or ndarry
+        Sky background.
+    fwhm : float or ndarray
+        PSF.
+    gain : float or ndarray; default is _ztfcam_ccd_gain
```
jeremykubica, line 76: Are there any instances where this wouldn't be `_ztfcam_ccd_gain` for ZTF? Since we are creating a subclass that just handles ZTF, I wondering if we will just always use `_ztfcam_ccd_gain`.

jeremykubica: Same question for the next two parameters.

jeremykubica (commented):
A few small comments.

src/tdastro/astro_utils/ztf_opsim.py
mi-dai: Not sure if this use case is actually real, but let's say the ccd gain (or the other parameters) measurement was off and we find it out after some time, we might want to simulate with a different value and check if this makes any difference in our science analysis.

src/tdastro/astro_utils/ztf_opsim.py
mi-dai: I added units for those

jeremykubica (approved):

PR #251, opened 2025-02-26
Create a new overview introductory notebook

jeremykubica:
Also renames the old `introduction_demo` notebook to make it clear that that is no longer the recommended starting point.


OliviaLynn (approved):
Looks good!

PR #252, opened 2025-02-26
Add initial citations

jeremykubica:
Add citation-compass as a dependency and use it to annotate some of the needed citations. We will need to do a more thorough search through the code base to make sure we are citing everything that we need to cite.


jeremykubica:
I updated citation-compass to handle that format and updated the comments in TDAstro.

hombit (approved):
Looks great, thank you

PR #253, opened 2025-02-27
Add a basic sampler to do (RA, dec) sampling from an opsim

jeremykubica:
Add a placeholder sampler to sample uniformly by the region covered by an OpSim. This uses rejection sampling and will be infeasibly slow for OpSims with small coverage. This is primarily a placeholder until we can add an approach that does some pre-computation of the OpSim region (which will be in a later PR).

### DIFF
--- a/src/tdastro/math_nodes/ra_dec_sampler.py
+++ b/src/tdastro/math_nodes/ra_dec_sampler.py
@@ -140,3 +140,101 @@ class OpSimRADECSampler(TableSampler):
             self._save_results(results, graph_state)
 
         return results
+
+
+class OpSimUniformRADECSampler(NumpyRandomFunc):
+    """A FunctionNode that samples RA and dec uniformly from the area covered
+    by an OpSim.  RA and dec are returned in degrees.
+
+    TODO: Implement a more efficient sampling method that precompiles the
+    opsim coverage and samples from that.
+
+    Note
+    ----
+    * This is a place holder class until we can implement a more efficient
+      sampling method (with precomputation of the opsim coverage).
+    * This uses rejection sampling and can be quite slow for small coverage.
+    * The sample will terminate after a maximum number of iterations to
+      prevent infinite loops. Some out of coverage samples may be returned.
+
+    Attributes
+    ----------
+    data : OpSim
+        The OpSim object to use for sampling.
+    radius : float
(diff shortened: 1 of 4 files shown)


src/tdastro/math_nodes/ra_dec_sampler.py
```diff
@@ -140,3 +140,101 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+        iter_num = 1
+        while num_missing > 0 and iter_num < 1000:
+            # Generate new samples for the missing ones.
+            ra[~mask] = np.degrees(rng.uniform(0.0, 2.0 * np.pi, size=num_missing))
+            dec[~mask] = np.degrees(np.arcsin(rng.uniform(-1.0, 1.0, size=num_missing)))
+
+            # Check if the samples are within the OpSim coverage.
+            mask = np.asarray(self.data.is_observed(ra, dec, self.radius))
+            num_missing = np.sum(~mask)
+            iter_num += 1
```
hombit, line 229: May I propose reducing code duplication with the initial mask being `np.zeros(ra.shape, dtype=bool)` and sampler running inside the loop?

src/tdastro/math_nodes/ra_dec_sampler.py
```diff
@@ -140,3 +140,101 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+        # If we are generating a single sample, return floats.
+        if graph_state.num_samples == 1:
+            ra = ra[0]
+            dec = dec[0]
+
+        # Set the outputs and return the results. This takes the place of
+        # function node's _save_results() function because we know the outputs.
+        graph_state.set(self.node_string, "ra", ra)
+        graph_state.set(self.node_string, "dec", dec)
+        return [ra, dec]
```
hombit, line 240: Not sure about list here, should it be tuple of np array?

hombit (approved):

src/tdastro/math_nodes/ra_dec_sampler.py
jeremykubica: Done.

src/tdastro/math_nodes/ra_dec_sampler.py
jeremykubica: That works. I changes the docstring as well.

PR #258, opened 2025-03-03
Opsim cleanup

jeremykubica:
Move some of the settings to internal class variables to make subclassing a bit cleaner.


hombit (approved):
It looks good, thank you

PR #259, opened 2025-03-03
Add the ability to sample from an OpSim's coverage

jeremykubica:
This PR adds the ability for users to generate a healsparse coverage map from an `OpSim` and to sample (RA, dec) points from that coverage map. It also adds a notebook tutorial demonstrating the different approaches for sampling (RA, dec) points.

### DIFF
--- a/docs/notebooks/sampling_positions.ipynb
+++ b/docs/notebooks/sampling_positions.ipynb
@@ -113,7 +113,7 @@
     "\n",
     "### Sampling Coverage\n",
     "\n",
-    "If we instead would like to sample uniformly from the area covered by the survey, we have two options. The `OpSimUniformRADECSampler` uses rejection sampling to generate positions. This approach can be slow for surveys with small coverage. However, since it requires no pre-computation, it is a viable option for generating a small number samples from a survey with high coverage."
+    "If we instead would like to sample uniformly from the area covered by the survey, we have two options. The `OpSimUniformRADECSampler` uses rejection sampling to generate positions. This approach can be slow for surveys with small coverage. However, since it requires no pre-computation, it is a viable option for generating a small number of samples from a survey with high coverage."
    ]
   },
   {
--- a/docs/notebooks/sampling_positions.ipynb
+++ b/docs/notebooks/sampling_positions.ipynb
@@ -15,7 +15,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import healpy as hp\n",
     "import numpy as np\n",
     "from tdastro.math_nodes.ra_dec_sampler import (\n",
     "    CoverageMapRADECSampler,\n",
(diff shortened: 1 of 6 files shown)


docs/notebooks/sampling_positions.ipynb
```diff
@@ -0,0 +1,203 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we can see, the field centered in the Northern hemisphere is sampled significantly more than the one centered in the Southern hemisphere.\n",
+    "\n",
+    "### Sampling Coverage\n",
+    "\n",
+    "If we instead would like to sample uniformly from the area covered by the survey, we have two options. The `OpSimUniformRADECSampler` uses rejection sampling to generate positions. This approach can be slow for surveys with small coverage. However, since it requires no pre-computation, it is a viable option for generating a small number samples from a survey with high coverage."
```
OliviaLynn, line 116:
```suggestion
    "If we instead would like to sample uniformly from the area covered by the survey, we have two options. The `OpSimUniformRADECSampler` uses rejection sampling to generate positions. This approach can be slow for surveys with small coverage. However, since it requires no pre-computation, it is a viable option for generating a small number of samples from a survey with high coverage."
```

pyproject.toml
```diff
@@ -20,6 +20,8 @@ dependencies = [
     "astropy",
     "citation-compass>=0.0.3",
     "dust_extinction",
+    "healpy",
```
OliviaLynn, line 23: This raises a small alarm for me, because I know LSDB has been working to remove healpy dependency this past year. 

I believe in part it is due to installation difficulties across multiple platforms, but in part it's because [healpy has a GPL license](https://github.com/healpy/healpy?tab=GPL-2.0-1-ov-file). Is this alright for TDAstro?

src/tdastro/math_nodes/ra_dec_sampler.py
```diff
@@ -237,3 +236,96 @@ def compute(self, graph_state, rng_info=None, **kwargs):
         graph_state.set(self.node_string, "ra", ra)
         graph_state.set(self.node_string, "dec", dec)
         return (ra, dec)
+
+
+class CoverageMapRADECSampler(NumpyRandomFunc, CiteClass):
+    """A FunctionNode that samples RA and dec uniformly from the area covered
+    by a survey as provided by either a opsim of a healsparse coverage map.
```
OliviaLynn, line 243:
```suggestion
    by a survey as provided by either an opsim or a healsparse coverage map.
```

OliviaLynn (approved):
Looks good!

pyproject.toml
jeremykubica: Let me look into this a bit.

jeremykubica: I'm only using healpy for testing and plotting. I removed both of those and switched to cdshealpix.

PR #260, opened 2025-03-03
Add a simple effect to model random drops due to chip gaps

jeremykubica:
Based on similar approaches within sorcha.

### DIFF
--- a/src/tdastro/effects/drop_observations.py
+++ b/src/tdastro/effects/drop_observations.py
@@ -1,0 +1,58 @@
+import numpy as np
+
+from tdastro.effects.effect_model import EffectModel
+
+
+class DropEffect(EffectModel):
+    """An effect that drops the observation with a given probability.
+    Used to model items falling in the telescope's field of view, but
+    not on a CCD (e.g. chip gaps)
+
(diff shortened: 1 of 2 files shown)


src/tdastro/effects/drop_observations.py
```diff
@@ -0,0 +1,58 @@
...
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of flux densities after the effect is applied (in nJy).
+        """
+        if rng_info is None:
+            rng_info = np.random.default_rng()
+
+        # Generate a random number for each observation
+        samples = rng_info.uniform(size=flux_density.shape)
+        return np.where(samples < self.drop_probability, 0.0, flux_density)
```
mi-dai, line 58: I think we want to actually "drop" it, not setting it to 0, because for 0 flux you are still getting an observation (or noise), but for chip gap you just don't get anything. Also I think this effect should be on the passband level? (What does it mean to randomly drop 10% of the SED and then integrate them over passband?)

jeremykubica:
Closing until we decide how we want to best handle chip gaps.

PR #261, opened 2025-03-04
add sfdmap

mi-dai:
Add SFDMap subclass from https://github.com/AmpelAstro/sfdmap2


src/tdastro/astro_utils/dustmap.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 from astropy.coordinates import SkyCoord
 from citation_compass import CiteClass
+from sfdmap2 import sfdmap
```
jeremykubica, line 11: For the tests to pass, you will need to include `sfdmap2` in the pyprojects.toml dependencies section.

src/tdastro/astro_utils/dustmap.py
```diff
@@ -144,3 +145,50 @@ def query(self, coords):
...
         """
         return self.compute_ebv(coords.ra.deg, coords.dec.deg)
+
+
+class SFDMap(DustEBV):
+    """
+    A dustmap using the sfdmap2 package (https://github.com/AmpelAstro/sfdmap2)
+    """
+
+    def __init__(self, **kwargs):
```
jeremykubica, line 155: You will need to expand this to download the data files if they do not already exist. They should go into the `data/dustmaps` directory. The best way to handle this would be the pooch library. @OliviaLynn is looking at adding this library for passbands, so now might be a good time to incorporate it throughout TDAstro.

src/tdastro/astro_utils/dustmap.py
```diff
@@ -144,3 +145,50 @@ def query(self, coords):
             The E(B-V) value or array of values.
         """
         return self.compute_ebv(coords.ra.deg, coords.dec.deg)
+
+
+class SFDMap(DustEBV):
+    """
+    A dustmap using the sfdmap2 package (https://github.com/AmpelAstro/sfdmap2)
+    """
+
```
jeremykubica, line 154: I would recommend adding a class variable here: `_default_map_dir= _TDASTRO_BASE_DATA_DIR / "sfdmap2"`

src/tdastro/astro_utils/dustmap.py
```diff
@@ -144,3 +145,50 @@ def query(self, coords):
...
+        dec : float or np.array
+            The object's declination (in degrees).
+
+        Returns
+        -------
+        ebv : float or np.array
+            The E(B-V) value or array of values.
+        """
+
+        dustmap = sfdmap.SFDMap()
```
jeremykubica, line 174: This should go in the constructor. Something like:

```
self.data_dir = self. _default_map_dir is data_dir is None else data_dir
self.dustmap = sfdmap.SFDMap(self.data_dir)
```

That way it is only loaded once.

jeremykubica (approved):

PR #262, opened 2025-03-05
Fill out the SFDMap node

jeremykubica:
Adds:
- The ability to download the data automatically if it does not exist on the local system
- Test dustmap files (really small files with constants for the Northern and Southern hemisphere).
- Tests for the new node.


mi-dai (approved):
Looks good! Thank you!

PR #266, opened 2025-03-06
build kdtree at last in case assign zero points changes the number of rows in the table

mi-dai:


mi-dai:
I'm very confused why the test is failing now but not before

jeremykubica:
This is failing because the test is returning the wrong type of error. Before it would hit a `KeyError` when looking for ra and dec to build the KD-tree. Now it hits a `ValueError` when trying to compute the zero points, because that happens first. I sent you a PR to fix your branch.

jeremykubica (approved):

PR #267, opened 2025-03-06
Add a bunch more comments to dustmaps.py

jeremykubica:
Add comments and parameters to clarify how the various models in dustmaps.py can be used, including:
- Include RA and dec in the parameter list so the user knows they need to be set when building a dust map node.
- Add a check that RA and dec have non-None setters and a comprehensible error message (before it was failing with a `TypeError` buried deep within the code)
- Make `ConstantHemisphereDustMap` just a dust map to be used with `DustmapWrapper` instead of letting it be a parameterized node and a dust map.


mi-dai (approved):
Looks good!

PR #274, opened 2025-03-24
add colormap to plotting.py; add redshift in the output table

mi-dai:


src/tdastro/simulate.py
```diff
@@ -90,6 +91,7 @@ def simulate_lightcurves(source, num_samples, opsim, passbands, rng=None):
         results_dict["ra"].append(ra[idx])
         results_dict["dec"].append(dec[idx])
         results_dict["nobs"].append(len(obs_times))
+        results_dict["z"].append(state["host.redshift"])
```
jeremykubica, line 94: I think this will need to be:
```
if "host.redshift" in state:
    results_dict["z"].append(state["host.redshift"])
```
In case the user passes a source with no host (or a host with a different name).

src/tdastro/utils/plotting.py
```diff
@@ -34,6 +28,8 @@ def plot_lightcurves(
         Figure, None by default.
     title : str or None, optional
         Title of the plot. None by default.
+    colormap: dict
```
jeremykubica, line 31: "colormap: dict" -> "colormap: dict, optional"

jeremykubica (changes requested):
A few small changes

src/tdastro/simulate.py
mi-dai: Is redshift a required parameter in source? Does this always work then? 
```z=source.get_param(sample_states, "redshift")
results_dict["z"].append(z[idx])```

Comment (jeremykubica) [reply src/tdastro/simulate.py:94]: It depends on what you mean by "required parameter". Every source will have a "redshift" parameter, but it might be set to None. So a modification of your suggestion will work:
```
results_dict["z"].append(source.get_param(state, "redshift"))
```
This will append either a redshift value or None if it is not set. Since it is using `source.get_params` (as with your example), you don't need to know the node's label.  So that should do what you want.

jeremykubica (approved):

PR #276, opened 2025-04-06
Provide basic extrapolation functions for sncosmo

jeremykubica:
Allow custom extrapolation beyond the wavelength bounds for sncosmo models. Since we might want to support this with other model types, we add an wave extrapolation class that can be reused (instead of putting this all in the sncosmo model itself).


mi-dai (approved):
Looks good!

PR #283, opened 2025-04-10
Create a source model that mimics a lightcurve

### LINKED ISSUE #282: Convert multiband light curve to an SED model
We should find a way to convert multiband light curves with passband transmissions given to an SED model

jeremykubica:
closes #282 

Creates a new `LightcurveSource` that converts per-band lightcurves into an SED model. Operations are done at the SED level, which is more expensive, but allows TDAstro to use all the effect infrastructure as-is.

### DIFF
--- a/src/tdastro/sources/lightcurve_source.py
+++ b/src/tdastro/sources/lightcurve_source.py
@@ -9,7 +9,7 @@ from tdastro.consts import lsst_filter_plot_colors
 from tdastro.sources.physical_model import PhysicalModel
 
 
-class LightcurveModel(PhysicalModel):
+class LightcurveSource(PhysicalModel):
     """A model that generates the SED of a source from lightcurves in given bands.
     The model estimates a box-shaped SED for each filter such that the resulting
     flux density is equal to the lightcurve's value after passing through
@@ -188,7 +188,7 @@ class LightcurveModel(PhysicalModel):
 
             # The contribution of this filter to the overall SED is the lightcurve's (interpolated)
             # value at each time multiplied by the SED values at each query wavelength.
-            sed_flux = np.outer(sed_waves, sed_time_mult)
+            sed_flux = np.outer(sed_time_mult, sed_waves)
             flux_density += sed_flux
 
         # Return the total flux density from all lightcurves.
--- a/src/tdastro/sources/lightcurve_source.py
+++ b/src/tdastro/sources/lightcurve_source.py
@@ -1,0 +1,262 @@
+"""A model that generates the SED of a source based on the lightcurves of fluxes
+in each band."""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from tdastro.astro_utils.passbands import Passband, PassbandGroup
+from tdastro.consts import lsst_filter_plot_colors
+from tdastro.sources.physical_model import PhysicalModel
+
+
+class LightcurveModel(PhysicalModel):
+    """A model that generates the SED of a source from lightcurves in given bands.
+    The model estimates a box-shaped SED for each filter such that the resulting
+    flux density is equal to the lightcurve's value after passing through
+    the passband filter.
+
+    The set of passbands used to configure the model MUST be the same as used
+    to generate the SED (the wavelengths must match).
+
+    Parameterized values include:
+      * dec - The object's declination in degrees.
+      * ra - The object's right ascension in degrees.
+      * t0 - The t0 of the zero phase (if applicable), date.
+
+    Attributes
+    ----------
+    lightcurves : dict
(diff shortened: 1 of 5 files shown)


src/tdastro/sources/lightcurve_source.py
```diff
@@ -0,0 +1,262 @@
...
+      * dec - The object's declination in degrees.
+      * ra - The object's right ascension in degrees.
+      * t0 - The t0 of the zero phase (if applicable), date.
+
+    Attributes
+    ----------
+    lightcurves : dict
+        A dictionary mapping filters to a 2d array representing the lightcurve
+        of the object. The first dimension is time and the second dimension
+        is the flux density in nJy.  An optional third dimension can be used to
```
mi-dai, line 31: Can we make this a table (ndarry) instead of a dict? I think it is more common to store light curves as tables than dicts.
Also the second dimension should be bandfluxes, not flux density.

src/tdastro/sources/lightcurve_source.py
```diff
@@ -0,0 +1,268 @@
...
+            # Get all of the wavelengths that have a non-negligible transmission value
+            # for this filter and find their indices in the passband group.
+            is_significant = passbands[filter].processed_transmission_table[:, 1] > 0.001
+            significant_waves = passbands[filter].waves[is_significant]
+            indices = np.searchsorted(passbands.waves, significant_waves)
+
+            # Mark all non-negligible wavelengths as used by this filter.
+            filter_uses_wave[idx, indices] = 1.0
+
+        # Find which wavelenfths are used by multiple filters.
```
mi-dai, line 117:
```suggestion
        # Find which wavelengths are used by multiple filters.
```

src/tdastro/sources/lightcurve_source.py
```diff
@@ -0,0 +1,268 @@
...
+        graph_state : GraphState
+            An object mapping graph parameters to their values.
+
+        Returns
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of rest frame SED values (in nJy).
+        """
+        params = self.get_local_params(graph_state)
+        shifted_times = times - params["t0"]
```
mi-dai, line 173: what happened if t0 is a random parameter here?

mi-dai (changes requested):

src/tdastro/sources/lightcurve_source.py
jeremykubica: If we make lightcurves an ndarray, we would need all the times to line up (and have the same number of samples).  Would we want to make that restriction?

src/tdastro/sources/lightcurve_source.py
jeremykubica: It will extract the sampled t0 for each run.  The `evaluate()` function loops over different samples, so by the time we get to this point we only have a single sample. The run will produce fluxes with the t0 for that sample.

src/tdastro/sources/lightcurve_source.py
mi-dai: I was thinking [[time1,time2,time3...],[flux1,flux2,flux3...],[filter1,filter2,filter3...]], and we don't need times to line up. so basically stack all times and fluxes together with the filter of each pair as the third dimension

jeremykubica: Sure. That's doable. I will make that change soon.

jeremykubica: I've added this as another configuration option. It will auto detect whether it is getting a dictionary or numpy array and process accordingly.

src/tdastro/sources/lightcurve_source.py
jeremykubica: I added the lightcurve specific t0 we discussed in slack.  Let me know if it what you were thinking.

src/tdastro/sources/lightcurve_source.py
```diff
@@ -39,19 +38,25 @@ class LightcurveSource(PhysicalModel):
...
+        The lightcurves can be passed as either:
+        1) a dictionary mapping filter names to a (T, 2) array of the bandlfuxes in that filter
+        where the first column is time and the second column is the flux density (in nJy), or
+        2) a numpy array of shape (T, 3) array where the first column is time (in days), the
+        second column is the bandflux (in nJy), and the third column is the filter.
     passbands : Passband or PassbandGroup
         The passband or passband group to use for defining the lightcurve.
+    lc_t0 : float
+        The zero time in the lightcurve arrays.  When computing flux the model's t0 will be
+        aligned with the lightcurve's lc_t0.
```
mi-dai, line 51: The reference epoch (t0) of the input light curve. The model will be shifted to the model's t0 when computing fluxes.

PR #286, opened 2025-04-10
Fix the benchmarking script

jeremykubica:
Benchmarks were failing because they couldn't find the passband files to load.


jeremykubica:
This doesn't fix the problem

jeremykubica:
Actually the benchmarking started working.

OliviaLynn (approved):

PR #289, opened 2025-04-14
Replace PhysicalModel's background with MultiSourceModel

### LINKED ISSUE #281: Model class that does linear combination of models
A model class that returns a linear combination of the TimeSeriesModel in #280 

```
class LinearCombModels(PhysicalModel):
    def __init__(weights, models):

    def compute_flux():
           flux = sum(weight[i]*model[i].flux)
```

jeremykubica:
Closes #281 

Add a `MultiSourceModel` that computes the flux as the (weighted) sum from a list of individual sources. Remove `PhysicalModel`'s background attribute in favor of this (**warning: this will be a breaking change**). This provides more flexibility and also consistency in how the effects are applied.

For testing purposes we add a toy effect `ConstantDimming`.

### DIFF
--- a/src/tdastro/effects/basic_effects.py
+++ b/src/tdastro/effects/basic_effects.py
@@ -1,0 +1,60 @@
+"""A collection of toy effect models that are primarily used for testing."""
+
+from tdastro.effects.effect_model import EffectModel
+
+
+class ConstantDimming(EffectModel):
+    """An effect that dims by a constant amount. Primarily used for testing.
+
+    Attributes
+    ----------
+    flux_fraction : parameter
+        The fraction of flux that is passed through.
+    rest_frame : bool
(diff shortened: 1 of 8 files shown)


src/tdastro/effects/basic_effects.py
```diff
@@ -0,0 +1,60 @@
+"""A collection of toy effect models that are primarily used for testing."""
+
+from tdastro.effects.effect_model import EffectModel
+
+
+class ConstantDimming(EffectModel):
```
hombit, line 6: `GreyFilterEffect`/`GreyDimmingEffect` are alternative names

hombit (approved):
Code looks good! I’m still not 100% sure whether we gain much from this approach compared to simply summing the resulting band-fluxes.

PR #297, opened 2025-05-20
Expand pzflow nodes to handle conditional models

jeremykubica:
Also moves the pzflow information into a (much expanded) stand alone notebook.


docs/notebooks/using_pzflow.ipynb
```diff
@@ -0,0 +1,294 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from tdastro.math_nodes.np_random import NumpyRandomFunc\n",
+    "\n",
+    "redshift_node = NumpyRandomFunc(\n",
+    "    \"uniform\",\n",
+    "    low=0.05,\n",
+    "    high=0.3,\n",
```
mi-dai, line 241: I suggest using the same range as before (0.01,0.25) so we can easily see the plots look similar

docs/notebooks/using_pzflow.ipynb
```diff
@@ -0,0 +1,294 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "state = source.sample_parameters(num_samples=100)\n",
+    "\n",
+    "plt.scatter(state[\"source\"][\"brightness\"], state[\"source\"][\"redshift\"], s=1)\n",
```
mi-dai, line 140: x, y are flipped.

mi-dai (changes requested):

docs/notebooks/using_pzflow.ipynb
jeremykubica: Fair. I had been using a different range intentionally to show how we would do that, but I think it is clearer with your suggestion.

docs/notebooks/using_pzflow.ipynb
jeremykubica: Fixed

PR #301, opened 2025-06-05
Add OpSim Download Functionality

### LINKED ISSUE #300: Automatic download OpSim File
Provide a mechanism to automatically download and use the OpSim file for LSST.

jeremykubica:
Closes #300 

Adds a helper function to make it easier to download OpSim databases. In the process, I broke out and generalized the passband download function.

### DIFF
--- a/docs/notebooks/opsim_notebook.ipynb
+++ b/docs/notebooks/opsim_notebook.ipynb
@@ -79,7 +79,28 @@
    "source": [
     "## Loading an OpSim\n",
     "\n",
-    "Most users will want to load a pre-existing `OpSim` from a database file using the `from_db()` function. "
+    "Most users will want to load a pre-existing `OpSim` from a database file using the `from_db()` function or `from_url()` functions.  The `from_url()` function will download the opsim db from a given URL to TDAstro's data cache (if needed) and load the opsim from a file.  For Rubin, a large number of OpSims can be found at [https://s3df.slac.stanford.edu/data/rubin/sim-data/](https://s3df.slac.stanford.edu/data/rubin/sim-data/). For many studies you will want the most recent version of a baseline survey. But some users might want older versions or alternate approaches."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "opsim_url = \"https://s3df.slac.stanford.edu/data/rubin/sim-data/sims_featureScheduler_runs3.4/baseline/baseline_v3.4_10yrs.db\"\n",
+    "\n",
+    "# Uncomment the following line to load data from a URL\n",
+    "# opsim_data = OpSim.from_url(opsim_url)"
+   ]
+  },
+  {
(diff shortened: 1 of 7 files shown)


docs/notebooks/opsim_notebook.ipynb
```diff
@@ -141,7 +162,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can run the spatial search in batch mode by providing a lists of RA and dec. The `range_search()` will return a list of numpy arrays where each element in the top-level list represents the matches for a single query (RA, dec). "
+    "We can run the spatial search in batch mode by providing a lists of RA and dec. The `range_search()` will return a list of numpy arrays where each element in the top-level list represents the matches for a single query (RA, Dec). "
```
OliviaLynn, line 165:
```suggestion
    "We can run the spatial search in batch mode by providing a lists of RA and Dec. The `range_search()` will return a list of numpy arrays where each element in the top-level list represents the matches for a single query (RA, Dec). "
```

docs/notebooks/opsim_notebook.ipynb
```diff
@@ -204,7 +225,7 @@
    "source": [
     "## Filtering\n",
     "\n",
-    "Depending on the analysis we run, we might want to need to use only a subset of the data.  The `OpSim` class provides a helper function `filter_rows()` that can be used to filter the data in an `OpSim` and thus speed up subsequent operations. This function creates a new copy of the data set with a subset of the rows."
+    "Depending on the analysis we run, we might want to need to use only a subset of the data.  `OpSim` provides a helper function `filter_rows()` that can be used to filter the data in an `OpSim` and thus speed up subsequent operations. This function creates a new copy of the data set with a subset of the rows."
```
OliviaLynn, line 228:
```suggestion
    "Depending on the analysis we run, we might want to use only a subset of the data.  `OpSim` provides a helper function `filter_rows()` that can be used to filter the data in an `OpSim` and thus speed up subsequent operations. This function creates a new copy of the data set with a subset of the rows."
```

docs/notebooks/opsim_notebook.ipynb
```diff
@@ -294,7 +315,7 @@
    "source": [
     "## Sampling\n",
     "\n",
-    "We can sample RA, dec, and time from an `OpSim` object using the `OpSimRADECSampler` node. This will select a random observation from the OpSim and then create a random (RA, dec) from near the center of that observation.\n",
+    "We can sample RA, dec, and time from an `OpSim` object using the `OpSimRADECSampler` node. This will select a random observation from the OpSim and then create a random (RA, Dec) from near the center of that observation.\n",
```
OliviaLynn, line 318:
```suggestion
    "We can sample RA, Dec, and time from an `OpSim` object using the `OpSimRADECSampler` node. This will select a random observation from the OpSim and then create a random (RA, Dec) from near the center of that observation.\n",
```
Not too sure on the standard for capitalizing Dec, but perhaps here too

OliviaLynn (approved):
Looks good!

PR #304, opened 2025-06-09
Allow the GivenValueSampler to take weights

jeremykubica:
Expand the interface of `GivenValueSampler` to take a vector of weights and sample from that distribution.


hombit (approved):
Looks great, thank you

Issue #299, opened 2025-06-05
Simulate from template light curves


berniez:
Hello
I was wondering if there was a way to simulate a source given a series of "perfect" light curves and the parameters used to generate them, similar to SNANA LCLIB. I can see a sort of roundabout way of doing it by using a LightCurve source to generate the SEDs, and then use it light the SplineModel. Is there a more direct way?

On a side note, do I need the OpSim file or can TDAstro download it (like it does for the filters)? Using a similar construct (preset="LSST") does not seem to work.

Thank you very much for the very nice work you all have done.

berniez:
> Hi [@berniez](https://github.com/berniez) , could you let us know a little more about why you think the LightCurve source doesn't work for your purpose?

@mi-dai upon further inspection, the LightCurve source can more or less do what I want. The one thing that Is lacking (or maybe I messed up) is the possibility to give more light curves (corresponding to different parameters used to generate them) as part of the source (not just one per filter). That is, my source would be characterized by several light curves instead of an analytical form.

berniez:
Thanks @jeremykubica. The way SNANA works would be fine, but in principle the probability should be the same for all curves. The manual has a clear explanation of how it selects the specific curves to be simulated, and at least for my purposes that is sufficient

PR #305, opened 2025-06-09
Expand the MultiSourceModel

jeremykubica:
Expand multiple source models to include both additive and (new) random models. Since they both use a lot of the same functionality, we make `MultiSourceModel` a base class and create new `AdditiveMultiSourceModel` and `RandomMultiSourceModel` classes.  

The motivation behind this is #299 and being able to sample from different light curve-based models.

### DIFF
--- a/src/tdastro/sources/multi_source_model.py
+++ b/src/tdastro/sources/multi_source_model.py
@@ -1,16 +1,18 @@
-"""The MultiSourceModel provides a wrapper to combine multiple
-PhysicalModels into a single model.
+"""Multiple source models wrap multiple PhysicalModels, allowing the
+user to define such operations as additive models, where each source
+contributes to the total flux density, or random source models, where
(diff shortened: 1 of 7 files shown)


src/tdastro/sources/multi_source_model.py
```diff
@@ -153,6 +144,81 @@ def sample_parameters(self, given_args=None, num_samples=1, rng_info=None):
...
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of SED values (in nJy).
+        """
+        raise NotImplementedError
+
+
+class AdditiveMultiSourceModel(MultiSourceModel):
+    """An AdditiveMultiSourceModel computes the flux from multiple overlapping objects,
+    including (host, source pairs) or unresolved sources.
```
olivialynn, line 177: Is this formatting intentional?

PR #307, opened 2025-06-11
Refactor LightcurveSource

jeremykubica:
As a pre-step toward creating a scalable multiple light curve model, this PR breaks the SED basis function computation out of `LightcurveSource`. This will allow us to compute one set of basis functions for all light curves that we load an use (instead of recomputing and storing for each light curve). Since the SED basis functions derived from the list of filters and the passband curves, they should be the same to all curves.

Most of this PR is moving code from `LightcurveSource` to `FakeSEDBasis`. However we also: 1) add some checks that the wavelengths used are a subset of those used to create the basis and 2) we switch from interpolation to using the indices directly.  These are both no-op changes when the set of wavelengths from the passbands is used to create and query models.

### DIFF
--- a/src/tdastro/sources/lightcurve_source.py
+++ b/src/tdastro/sources/lightcurve_source.py
@@ -275,7 +275,6 @@ class LightcurveSource(PhysicalModel):
         self.lightcurves = {}
 
         # Store the lightcurve information.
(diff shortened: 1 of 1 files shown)


jeremykubica:
Closed because we went with a different approach.

PR #309, opened 2025-06-13
Add a note on downloading the OpSim data

jeremykubica:
Add a note about downloading the OpSim data to the plasticc_snia notebook.

### DIFF
--- a/docs/notebooks/pre_executed/plasticc_snia.ipynb
+++ b/docs/notebooks/pre_executed/plasticc_snia.ipynb
@@ -130,7 +130,7 @@
    ],
    "source": [
     "# Load the OpSim data.\n",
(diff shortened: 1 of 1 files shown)


aliceliu723 (changes requested):
The instruction looks good! I noticed in the **plasticc_snia.ipynb**, the path is "_TDASTRO_BASE_DATA_DIR / "opsim_db" / "baseline_v3.4_10yrs.db". However, in the **from_url()** helper function, the path is "_TDASTRO_BASE_DATA_DIR / "opsim" / data_file_name". One is **"opsim_db"**, the other one is just **"opsim"**. I don't know which one is the one we want to actually download the file in, but it would be good to make them consistent I think.

jeremykubica:
Good catch. I think we want to use "opsim". I changed the notebook to be consistent.

aliceliu723 (approved):

PR #311, opened 2025-06-13
Break lightcurve data into its own object

jeremykubica:
Break out the lightcurve-based data from `LightcurveSource`. This is an initial step before allowing a source to have a list of lightcurves, in which case this will be a list of `LightcurveData`.

Most of this PR is moving functionality from `LightcurveSource` to `LightcurveData`.  It also adds a helper function to read `LightcurveData` from an LCLIB style table.

### DIFF
--- a/src/tdastro/sources/lightcurve_source.py
+++ b/src/tdastro/sources/lightcurve_source.py
@@ -5,7 +5,6 @@ import warnings
 
 import matplotlib.pyplot as plt
 import numpy as np
-from astropy.table import Table
 
 from tdastro.astro_utils.passbands import Passband, PassbandGroup
 from tdastro.consts import lsst_filter_plot_colors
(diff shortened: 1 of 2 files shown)


src/tdastro/sources/lightcurve_source.py
```diff
@@ -9,42 +11,20 @@
...
-
-    The set of passbands used to configure the model MUST be the same as used
-    to generate the SED (the wavelengths must match).
-
-    Parameterized values include:
-      * dec - The object's declination in degrees.
-      * ra - The object's right ascension in degrees.
-      * t0 - The t0 of the zero phase (if applicable), date.
+class LightcurveData:
+    """A class to hold data for a single lightcurve.
```
hombit, line 15:
```suggestion
    """A class to hold data for a single model lightcurve.
```

hombit (approved):
Looks good

PR #312, opened 2025-06-13
Break fake SED basis functions into their own class

jeremykubica:
This PR replaces #307 

As a pre-step toward creating a scalable multiple light curve model, this PR breaks the SED basis function computation out of LightcurveSource. This will allow us to compute one set of basis functions for all light curves that we load an use (instead of recomputing and storing for each light curve). Since the SED basis functions derived from the list of filters and the passband curves, they should be the same to all curves.

Most of this PR is moving code from LightcurveSource to FakeSEDBasis. However we also: 1) add some checks that the wavelengths used are a subset of those used to create the basis and 2) we switch from interpolation to using the indices directly. These are both no-op changes when the set of wavelengths from the passbands is used to create and query models.

### DIFF
--- a/src/tdastro/astro_utils/fake_sed_basis.py
+++ b/src/tdastro/astro_utils/fake_sed_basis.py
@@ -1,0 +1,169 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from tdastro.astro_utils.passbands import Passband, PassbandGroup
+from tdastro.consts import lsst_filter_plot_colors
+
+
+class FakeSEDBasis:
+    """A fake SED class that is used to create a box-shaped SED for each filter.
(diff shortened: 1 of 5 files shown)


mi-dai (commented):
I made a lot of suggestions. The main concern is we should be clear about what is SED, what is SED basis. For me the box shaped thing that is 1 over the integral of the passband values is SED basis, since it is multiplied later by the light curve values to create the SED, so we should avoid calling it SED. (Honestly, SED basis is also not very easy to understand, but I can't come up with a better name)

Some other variable names confused me a bit when I tried to understand what the code does.

tagging @hombit for a second look in case I'm being too annoying. Feel free to ignore my suggestion if they don't all make sense.

PR #316, opened 2025-06-16
Add support for the multiple light curve model

jeremykubica:
Add support for a `MultiLightcurveSource` that randomly selects a lightcurve and then evaluates it. To avoid code duplication, we break most of the code for `LightcurveSource` and `MultiLightcurveSource` into a new base class `BaseLightcurveSource`, which hands all the information about the basis SED functions, computing the flux, setting restrictions on applying redshift, etc. This makes the two source classes very thin wrappers on the base class.

### DIFF
--- a/src/tdastro/sources/lightcurve_source.py
+++ b/src/tdastro/sources/lightcurve_source.py
@@ -2,6 +2,7 @@
 in each band."""
 
 import warnings
+from abc import ABC
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/tdastro/sources/lightcurve_source.py
```diff
@@ -294,21 +296,11 @@ def plot_lightcurves(self, times=None, ax=None, figure=None):
...
-    light curve is not periodic then each lightcurve's given values will be interpolated
-    during the time range of the lightcurve. Values outside the time range (before and
-    after) will be set to the baseline value for that filter (0.0 by default).
-
-    Periodic models require that each filter's lightcurve is sampled at the same times
-    and that the value at the end of the lightcurve is equal to the value at the start
-    of the lightcurve. The lightcurve epoch (lc_t0) is automatically set to the first time
-    so that the t0 parameter corresponds to the shift in phase.
+class BaseLightcurveSource(PhysicalModel):
+    """A base class for lightcurve source models. This class is not meant to be used directly,
```
hombit, line 300: Maybe inherit from `abc.ABC`, so it may not be used directly?

hombit (approved):
Looks good

PR #321, opened 2025-06-17
Update PassbandGroup API

jeremykubica:
Update the `PassbandGroup` API based on suggestions from @mi-dai. The main change is to break out the presets from the constructor into its own `from_preset` function.

### DIFF
--- a/src/tdastro/astro_utils/passbands.py
+++ b/src/tdastro/astro_utils/passbands.py
@@ -45,7 +45,7 @@ class PassbandGroup:
 
         Parameters
         ----------
-        given_passbands : list of Passband or dict
+        given_passbands : Passband, list of Passband, or dict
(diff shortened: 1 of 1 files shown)


src/tdastro/astro_utils/passbands.py
```diff
@@ -87,53 +71,44 @@ def __init__(
...
-                "PassbandGroup must be initialized with one of a preset, a list"
-                "of Passband objects, or passband_parameters."
-            )
-
-        if preset is not None:
-            self.add_preset(preset, table_dir=table_dir, **kwargs)
+        # If we are given a single Passband object or a single dictionary, wrap it in a list.
+        if isinstance(given_passbands, Passband):
+            given_passbands = [given_passbands]
+        elif isinstance(given_passbands, dict):
```
olivialynn, line 77: .

PR #325, opened 2025-06-23
Allow lightcurves to compute bandfluxes directly

jeremykubica:
Instead of using a basis function to compute the bandpass at different wavelengths, we can just interpolate the bandflux for each filter when the user asks for bandflux level data. This both removes an approximation and in decreases the cost of the computation.


src/tdastro/sources/lightcurve_source.py
```diff
@@ -645,6 +719,38 @@ def compute_flux(self, times, wavelengths, graph_state):
...
+        params = self.get_local_params(state)
+
+        # Shift the times for the model's t0 aligned with the lightcurve's lc_t0.
+        # The lightcurve times were already shifted in the constructor to be relative to lc_t0.
+        shifted_times = times - params["t0"]
+
+        band_fluxes = np.zeros(len(times))
+        for filter in self.lightcurves.filters:
+            filter_mask = filters == filter
+            band_fluxes[filter_mask] = self.lightcurves.evaluate(shifted_times[filter_mask], filter)
```
mi-dai, line 750: This is doing simple interpolation of the light curves, and has nothing to do with sed, right? I got quite confused first looking at [L257](https://github.com/lincc-frameworks/tdastro/blob/cc2ed1ba6c48a078756190234e404084cd07a3d1/src/tdastro/sources/lightcurve_source.py#L257) because it describes the light curve as sed.

mi-dai (approved):

Issue #330, opened 2025-06-27
Output template mags when using LCLIB files


berniez:
**Feature request**
Hi
I thought I'd add this as a separate request, but it's related to #299. Some non-recurring non-periodical events, such as stellar flares, have a baseline magnitude value. This is important to have on the output, since it's needed to establish if any given point counts as a detection (5 sigma above the baseline, which is not the 5sigma depth anymore). For an LCLIB file these are the rows that start with "T:", but it can be given by the user.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #331, opened 2025-06-27
Add a function node to select a given value from a list

jeremykubica:
Create a node which takes an index as its parameter and returns the corresponding value from a list. This node will be needed to address issue #330 by storing the templated parameters as lists.

### DIFF
--- a/tests/tdastro/math_nodes/test_given_sampler.py
+++ b/tests/tdastro/math_nodes/test_given_sampler.py
@@ -118,7 +118,7 @@ def test_given_value_selector():
     index_node = GivenValueList([0, 1, 2, 3, 2, 3, 1, 2], node_label="index_node")
     given_node = GivenValueSelector([10, 20, 30, 40], index_node, node_label="given_node")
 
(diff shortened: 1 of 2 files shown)


tests/tdastro/math_nodes/test_given_sampler.py
```diff
@@ -108,6 +113,20 @@ def test_given_value_sampler():
     assert len(results[results == 7]) > 1000
 
 
+def test_given_value_selector():
+    """Test that we can retrieve numbers from a GivenValueSelector."""
+    index_node = GivenValueList([0, 1, 2, 3, 2, 3, 1, 2], node_label="index_node")
+    given_node = GivenValueSelector([10, 20, 30, 40], index_node, node_label="given_node")
+
+    # Check that we have sampled uniformly from the given options.
```
olivialynn, line 121: The wording here is a little confusing to me—we're just sampling based on the list we've given it here, right?

olivialynn (approved):

PR #337, opened 2025-07-14
Add check that LightcurveSource models support requested filters

jeremykubica:
Fail with a clear error if querying the bandflux for a LightcurveSource model that does not have that filter. This prevents the model from returning zero without any warning.


jeremykubica:
I removed that test because it was checking that the model filled in zeros if the filter was not supported, which is not what we want.

PR #344, opened 2025-07-18
modified salt2jax model 

mi-dai:
`SALT2JaxModel`'s `compute_flux()` was taking phase instead of times and was giving wrong values. updated to calculate phase from times, and updated tests to use non-zero t0. (Old tests did not fail because t0=0 so phase=times)


jeremykubica (approved):
Thank you!

PR #352, opened 2025-07-24
Fix issue 351

### LINKED ISSUE #351: Implement cos(i) distribution for the AGN model
**Bug report**

AGN model always assume i=0 deg, we should sample i for the uniform distribution [0; 90] deg

https://github.com/lincc-frameworks/tdastro/blob/864ea3b82438b92ff592260da7a144ef4b2547a2/src/tdastro/sources/agn.py#L439

**Before submitting**
Please check the following:

- [ ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.

jeremykubica:
Closes #351 

Sample's the AGN's inclination uniformly from (0, pi/2)

### DIFF
--- a/src/tdastro/sources/agn.py
+++ b/src/tdastro/sources/agn.py
@@ -13,6 +13,7 @@ from scipy import integrate
 
 from tdastro.base_models import FunctionNode
 from tdastro.consts import M_SUN_G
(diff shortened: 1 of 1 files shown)


hombit (approved):
Thank you!

PR #355, opened 2025-07-25
Create a Passband only model type

jeremykubica:
Create a new abstract base class for supporting passband only models. We base this off `BaseLightcurveSource` and move some of that class's functionality to the new base class.

In addition we add support for effects at the band flux level. Not all effects will be supported, but this provides the ability to better run simulations with band flux only models.


hombit (approved):
Thank you, I really like it! @mi-dai and myself discussed naming and we propose to change `PassbandModel -> BandfluxModel` and open an issue and change `PhysicalModel -> SEDModel` after the hack.

PR #364, opened 2025-07-28
Create a model that chooses a random SED to use

### LINKED ISSUE #360: Static SED Model
We want to make a model that reads in an SED with wavelengths and values, and then we query that and interpolates at the wavelengths we don't have.

Use linear interpolation.
Doesn't change over time.

Start with:
- Static SEDs over time
- Sample from a library of SEDs (and pick randomly from them)
- Don't worry how the data gets in; assuming someone compiles the SED files for you
- We can figure out how people have/will configured such files, but we like the idea of supporting NestedFrames to start. 
- For now, pick some format—maybe a list of numpy arrays

jeremykubica:
Closes #360 

Uses that SED as a constant over all time steps.

### DIFF
--- a/src/tdastro/sources/multi_source_model.py
+++ b/src/tdastro/sources/multi_source_model.py
@@ -225,7 +225,7 @@ class AdditiveMultiSourceModel(MultiSourceModel):
 
         Note
         ----
-        Wavelength extrapolation is handled by each source.  So the actual wavelength's
+        Wavelength extrapolation is handled by each source. So the actual wavelength's
         can be evaluated outside the range of each source.
 
         Parameters


src/tdastro/sources/static_sed_source.py
```diff
@@ -0,0 +1,105 @@
...
+            A length N array of observer frame wavelengths (in angstroms).
+        graph_state : GraphState
+            An object mapping graph parameters to their values.
+
+        Returns
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of observer frame SED values (in nJy). These are generated
+            from non-overlapping box-shaped SED basis functions for each filter and
+            scaled by the lightcurve values.
```
mi-dai, line 87: This last sentence about box-shaped SED is not applicable here.

src/tdastro/sources/static_sed_source.py
```diff
@@ -0,0 +1,105 @@
...
+        num_waves = len(wavelengths)
+
+        # At each time step we interpolate SED at the query wavelengths.
+        sed_fluxes = np.interp(
+            wavelengths,
+            self.sed_values[model_ind][0, :],
+            self.sed_values[model_ind][1, :],
+            left=0.0,
+            right=0.0,
+        )
```
mi-dai, line 101: Will there be any extrapolation handled by PhysicalModel or should extrapolation be enabled here?

src/tdastro/sources/static_sed_source.py
jeremykubica: Fixed.

src/tdastro/sources/multi_source_model.py
```diff
@@ -219,6 +219,42 @@ def __init__(
         else:
             self.weights = weights
 
+    def minwave(self, graph_state=None):
+        """Get the minimum wavelength of the model. For additive models, this is
+        the minimum wavelength of all sources.
```
mi-dai, line 224: I'm not sure about this. This may cause some artifacts if one of the model is truncated instead of physically having shorter coverage. If we'd like to use this approach, I would do the maximum of all minwaves for minwave and minimum of all maxwaves for maxwave. Or we should do extrapolations for each source indivicually so they end up having the same min/maxwaves.

mi-dai: This sounds great! But is this min/maxwave() in additive model used anywhere then?

mi-dai (approved):

PR #374, opened 2025-07-29
Add helpers for recreating a GraphState

### LINKED ISSUE #371: Add function for creating GraphState from dictionary
Allow a `GraphState` to be reconstructed from the output param's columns. This should allow it either to take single dictionary or an iterable of dictionaries.

jeremykubica:
Closes #371 

Add helpers to recreate a `GraphState` object from the dictionaries that we store in the output table.


olivialynn (approved):
Looks good!

PR #376, opened 2025-08-04
Rename core classes and functions

### LINKED ISSUE #357: Rename PhysicalModel to SEDModel
After the LINCC FW hack days, let's rename `PhysicalModel` to `SEDModel`. It would help to understand how `SEDModel` and `BandfluxModel` are different.

jeremykubica:
Closes #357

This PR is a relatively major renaming of the core classes and functions to improve consistency: `PhysicalModel` is split into `BasePhysicalModel` and `SEDModel` (a subclass). `BasePhysicalModel` contains all the logic that is common to both SED-type models and bandflux-type models. While `SEDModel` contains all the SED-handling functionality. The goal of this split is to avoid bandflux-type models having sed-level logic that can erroneously be triggered.


mi-dai:
I find these renaming a bit confusing. Do we expect to use anything else other than `evaluate` to calculate the flux density matrix? What term is used for light curve model currently? The original distinction I had in mind is `evaluate` is directly calculating flux density defined by the model, while `get_band_fluxes` requires to define a passband and do integration (and that's why we don't call it evaluate something like the seds because it can't be done with the model itself)

docs/notebooks/adding_sources.ipynb
```diff
@@ -10,14 +10,22 @@
     "\n",
     "## Creating a new model node\n",
     "\n",
-    "All sources are subclasses of the `PhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object, but users only need to override two of them to create a new subclass:\n",
+    "All sources are subclasses of the `BasePhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object. Users should not subclass  `BasePhysicalModel` directly, but rather subclass either `SEDModel` (which generates SED-level outputs) or `BandfluxModel`\n",
+    "(which generates blandflux-level outputs). Depending on the model type, users only need to override a small number of functions.\n",
```
olivialynn, line 14:
```suggestion
    "All sources are subclasses of the `BasePhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object. Users should not subclass `BasePhysicalModel` directly, but rather subclass either `SEDModel` (which generates SED-level outputs) or `BandfluxModel`\n",
    "(which generates bandflux-level outputs). Depending on the model type, users only need to override a small number of functions.\n",
```

docs/notebooks/adding_sources.ipynb
```diff
@@ -109,15 +117,17 @@
...
     "\n",
-    "**Note:** TDAstro is set up to be modular with the foundation code performing most of the heavy listing. This includes most of the functions in `PhysicalModel`, because the base class is set up to consistently apply transformations and effects to all its subclasses.  New effects can be added by creating new subclasses of `EffectModel`. Thus the developers of new sources should avoid overloading or modifying the `evaluate()` function unless they need to fundamentally change how effects are applied."
+    "**Note:** TDAstro is set up to be modular with the foundation code performing most of the heavy listing. This includes most of the functions in `SEDModel` and `BandfluxModel`, because these base classes are set up to consistently apply transformations and effects to all its subclasses.  New effects can be added by creating new subclasses of `EffectModel`. Thus the developers of new sources should avoid overloading or modifying the `evaluate_` functions unless they need to fundamentally change how effects are applied."
```
olivialynn, line 130:
```suggestion
    "**Note:** TDAstro is set up to be modular with the foundation code performing most of the heavy lifting. This includes most of the functions in `SEDModel` and `BandfluxModel`, because these base classes are set up to consistently apply transformations and effects to all their subclasses.  New effects can be added by creating new subclasses of `EffectModel`. Thus the developers of new sources should avoid overloading or modifying the `evaluate_` functions unless they need to fundamentally change how effects are applied."
```

docs/notebooks/introduction.ipynb
```diff
@@ -102,7 +102,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The power of the simulation software is that we can generate a large number of light curves from a distribution of sources. We start by using a `PhysicalModel` object's `sample_parameters` function to sample the parameters can create this distribution of source object. \n",
+    "The power of the simulation software is that we can generate a large number of light curves from a distribution of sources. We start by using a `BasePhysicalModel` object's `sample_parameters` function to sample the parameters can create this distribution of source object. \n",
```
olivialynn, line 105:
```suggestion
    "The power of the simulation software is that we can generate a large number of light curves from a distribution of sources. We start by using a `BasePhysicalModel` object's `sample_parameters` function to sample the parameters that can create this distribution of source object. \n",
```

docs/notebooks/lightcurve_source_demo.ipynb
```diff
@@ -64,7 +62,7 @@
    "source": [
     "## Creating the model\n",
     "\n",
-    "We create the model from lightcurves for each passbands of interest. This is what we want the model to reproduce then we call get_band_fluxes(). While these lightcurves do no need to be the same as in the `PassbandGroup` every lightcurve must have a corresponding entry in the `PassbandGroup`.\n",
+    "We create the model from lightcurves for each passbands of interest. This is what we want the model to reproduce then we call `evaluate_bandflux()`. While these lightcurves do no need to be the same as in the `PassbandGroup` every lightcurve must have a corresponding entry in the `PassbandGroup`.\n",
```
olivialynn, line 65:
```suggestion
    "We create the model from lightcurves for each passbands of interest. This is what we want the model to reproduce then we call `evaluate_bandflux()`. While these lightcurves do not need to be the same as in the `PassbandGroup`, every lightcurve must have a corresponding entry in the `PassbandGroup`.\n",
```

src/tdastro/sources/galaxy_models.py
```diff
@@ -1,21 +1,21 @@
...
-      * distance - The object's luminosity distance in pc. [from PhysicalModel]
+      * dec - The object's declination in degrees. [from BasePhysicalModel]
+      * distance - The object's luminosity distance in pc. [from BasePhysicalModel]
       * galaxy_radius_std - The standard deviation of the brightness in degrees.
-      * ra - The object's right ascension in degrees. [from PhysicalModel]
-      * redshift - The object's redshift. [from PhysicalModel]
-      * t0 - No effect for a GuassianGalaxy [from PhysicalModel]
+      * ra - The object's right ascension in degrees. [from BasePhysicalModel]
+      * redshift - The object's redshift. [from BasePhysicalModel]
+      * t0 - No effect for a GuassianGalaxy [from BasePhysicalModel]
```
olivialynn, line 18:
```suggestion
      * t0 - No effect for a GuassianGalaxy. [from BasePhysicalModel]
```

src/tdastro/sources/physical_model.py
```diff
@@ -1,12 +1,12 @@
...
-A BandfluxModel is a PhysicalModel that only computes band fluxes instead of the
-SEDs. This is used for models that are emperically fit from observed band fluxes.
-We strongly recommend using the full SED models (PhysicalModel) whenever possible
-since they more accurately simulate aspects such as the impact of redshift on rest
-frame effects.
+"""The base classes for all sources.
+
+The code supports two types of models: 1) SEDModels define recipes for computing SEDs
+from a source at given times and wavelengths, accounting for redshift and other effects.
+2) BandfluxModels only computes band fluxes instead of the SEDs. This is used for models
```
olivialynn, line 5:
```suggestion
2) BandfluxModels only compute band fluxes instead of the SEDs. This is used for models
```

src/tdastro/sources/spline_model.py
```diff
@@ -22,11 +22,11 @@ class SplineModel(PhysicalModel):
...
-      * dec - The object's declination in degrees. [from PhysicalModel]
-      * distance - The object's luminosity distance in pc. [from PhysicalModel]
-      * ra - The object's right ascension in degrees. [from PhysicalModel]
-      * redshift - The object's redshift. [from PhysicalModel]
-      * t0 - The time corresponding to the t=0.0 in the model data [from PhysicalModel]
+      * dec - The object's declination in degrees. [from BasePhysicalModel]
+      * distance - The object's luminosity distance in pc. [from BasePhysicalModel]
+      * ra - The object's right ascension in degrees. [from BasePhysicalModel]
+      * redshift - The object's redshift. [from BasePhysicalModel]
+      * t0 - The time corresponding to the t=0.0 in the model data [from BasePhysicalModel]
```
olivialynn, line 29:
```suggestion
      * t0 - The time corresponding to the t=0.0 in the model data. [from BasePhysicalModel]
```

olivialynn (commented):
Looks good!

docs/glossary.rst
```diff
@@ -1,14 +1,18 @@
 Glossary
 ========================================================================================
 
-**BandfluxModel**: A ``BandfluxModel`` is a subclass of the ``PhysicalModel`` class that represents the flux density of a source in a specific band, such as the g-band or r-band (instead of a full spectral energy distribution).
+**BandfluxModel**: A ``BandfluxModel`` is a subclass of the ``BasePhysicalModel`` class that represents the flux density of a source in specific bands (band-fluxes), such as the g-band or r-band (instead of a full spectral energy distribution).
```
mi-dai, line 4: How about "represents broadband photometry in specific bands..." @hombit?

docs/notebooks/adding_sources.ipynb
```diff
@@ -10,14 +10,22 @@
     "\n",
     "## Creating a new model node\n",
     "\n",
-    "All sources are subclasses of the `PhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object, but users only need to override two of them to create a new subclass:\n",
+    "All sources are subclasses of the `BasePhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object. Users should not subclass `BasePhysicalModel` directly, but rather subclass either `SEDModel` (which generates temporal spectral outputs) or `BandfluxModel`\n",
```
mi-dai, line 13:
```suggestion
    "All sources are subclasses of the `BasePhysicalModel` class in `src/tdastro/sources/physical_models.py`. This class implements a bunch of helper functions to handle parameters and simulate the output of an object. Users should not use subclass `BasePhysicalModel` directly, but rather subclass either `SEDModel` (which generates temporal spectral outputs) or `BandfluxModel`\n",
```

src/tdastro/sources/multi_source_model.py
```diff
@@ -206,19 +209,19 @@ def _evaluate_band_fluxes_single(
 
 class AdditiveMultiSourceModel(MultiSourceModel):
     """An AdditiveMultiSourceModel computes the flux from multiple overlapping objects,
-    including (host, source pairs) or unresolved sources.
+    including (host, galaxy pairs) or unresolved sources.
```
mi-dai, line 212:
```suggestion
    including (host galaxy and source pairs) or unresolved sources.
```

docs/glossary.rst
hombit: @mi-dai I like it

PR #383, opened 2025-08-12
Create a simple binary sampler

jeremykubica:
Create a simple `BinarySampler` class to use for turning on/off effects. The idea is that we can include this into the graph when we want to randomly apply an effect to a fraction of the sources.


hombit (approved):
Looks great!

PR #385, opened 2025-08-12
Add Microlensing effect

jeremykubica:
Add a microlensing effect adapted from @KKruszynska's hackday code.

Co-authored-by: @KKruszynska


hombit (approved):
It is actually looks really good! I'm still not convinced that we need the probability parameter - for a random star in a field the probability would be pretty low and I'm not sure that we'd like to generate a lot of non-lensed star. Could the probability itself be a random number?

PR #388, opened 2025-08-13
Simplify BandfluxModel's get_band_fluxes

jeremykubica:
`BandfluxModel.get_band_fluxes()` does not use the `passband_or_group`. We just have it there to use the same API as the overall `PhysicalModel`. This PR makes it clear that it is optional and removes the fake PassbandGroup from a test to simplify it.

### DIFF
--- a/src/tdastro/sources/physical_model.py
+++ b/src/tdastro/sources/physical_model.py
@@ -14,7 +14,7 @@ from os import urandom
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


hombit (approved):
Looks good. This makes me think that we should use keyword-only arguments more often (aka `f(x, *, y, z=None)`), so the caller's code doesn't change much when we make these changes

PR #389, opened 2025-08-13
Add some debugging helpers for effects

jeremykubica:
Add some small debugging helper for effects, including:
- The ability to set an effect name.
- Human readable string and repr functions
- The ability to get all of the effects added to a source.

### DIFF
--- a/src/tdastro/effects/effect_model.py
+++ b/src/tdastro/effects/effect_model.py
@@ -46,7 +46,7 @@ class EffectModel:
         return self.effect_name
 
     def __repr__(self):
-        # We only return the parameter names, because the values may be dynamically
+        # We only include the parameter names because the values may be dynamically
         # sampled during the simulation.
         params = ",".join(self.parameters.keys())
         return f"{self.effect_name}({params})"


src/tdastro/effects/effect_model.py
```diff
@@ -20,19 +20,35 @@ class EffectModel:
...
+
         # Automatically include all keyword arguments as settable parameters.
         self.parameters = {}
         for key, value in kwargs.items():
             self.add_effect_parameter(key, value)
 
+    def __str__(self):
+        return self.effect_name
+
+    def __repr__(self):
```
olivialynn, line 48:
```suggestion
    def __repr__(self):
        # Return parameter name only (and not value) to avoid unnecessarily long strings when chaining nodes.
```

It took me a minute to reason about this, so it might be nice to add a note to explain that we've already considered the pros and cons of a more typical repr output

olivialynn (approved):
Looks good!

src/tdastro/effects/effect_model.py
jeremykubica: That is actually not the reason. Parameters can be dynamically sampled, so we do not know their value ahead of time.

jeremykubica: I added a comment about this.

PR #395, opened 2025-08-19
Make VBMicrolensing an optional dependency

jeremykubica:
VBMicrolensing is not currently available on conda-forge, so we make it an optional dependency and skip the its tests if it is not installed. Note that the github actions, including the precommit are updated to force its installation, so those github tests should always be run.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -74,6 +74,11 @@ You can install TDAstro from PyPI with pip or from conda-forge with conda. We re
           >> # Install from conda-forge channel
           >> conda install conda-forge::tdastro
 
+Since TDAstro relies on a large number of existing packages, not all of the packages
+are installed in the default configuration. For example the microlensing package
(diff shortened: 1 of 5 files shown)


hombit (commented):
Few opinionated comments:
1. I'd import `VBMicrolensing` inside the class uses it, instead of the top-level module import, so `tdastro`'s module import never fails. I do like the idea of the error message with `pip install` recommendation, but I'd move it to the class initialization. I believe that packages like  `astropy` do it this way
2. I'd remove checks in the tests, we should be sure that tests actually run.
3. Let's copy-paste `all_astro` deps to `dev`

hombit (approved):
Sorry, I didn't mean to remove `all-astro` deps, I just meant to have `VBMicrolensing` in both `dev` and `all-astro`. Let's put `all-astro` back (maybe rename it to just `all`?)?

I'm fine with merging this. We can add `VBMicrolensing` as a `conda-forge` dependency later when/if it becomes available there.

PR #396, opened 2025-08-19
Reorganize get_band_fluxes

jeremykubica:
Reorganize the `get_band_fluxes()` function so we can reuse more of the logic in `BandfluxModel`. This is a precursor change to expanding the support for various additive models.

Also allow effects to not propagate their parameters in case they are added via other mechanisms (also a precursor change).

### DIFF
--- a/src/tdastro/sources/physical_model.py
+++ b/src/tdastro/sources/physical_model.py
@@ -14,7 +14,7 @@ from os import urandom
 
 import numpy as np
 
-from tdastro.astro_utils.passbands import Passband
+from tdastro.astro_utils.passbands import Passband, PassbandGroup
 from tdastro.astro_utils.redshift import RedshiftDistFunc, obs_to_rest_times_waves, rest_to_obs_flux
 from tdastro.base_models import ParameterizedNode
 from tdastro.graph_state import GraphState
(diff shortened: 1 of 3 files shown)


tests/tdastro/effects/test_basic_effects.py
```diff
@@ -46,7 +46,10 @@ def test_static_source_constant_dimming() -> None:
     assert len(model.rest_frame_effects) == 1
     assert len(model.obs_frame_effects) == 0
 
+    # Check that the flux_fraction parameter is stores in the source node.
```
olivialynn, line 49:
```suggestion
    # Check that the flux_fraction parameter is stored in the source node.
```

src/tdastro/sources/physical_model.py
```diff
@@ -610,35 +665,20 @@ def get_band_fluxes(self, passband_or_group, times, filters, state, rng_info=Non
...
-                    **params,  # Provide all the node's parameters to the effect.
-                )
-            results[sample_num, :] = band_flux
+        params = self.get_local_params(state)
 
-        if state.num_samples == 1:
-            return results[0, :]
-        return results
+        # Compute the flux (applying all effects) and save the result. Note that
+        # BandfluxModel do not apply redshift, so all effects are applied in observer frame.
```
olivialynn, line 674:
```suggestion
        # BandfluxModel does not apply redshift, so all effects are applied in observer frame.
```

olivialynn (approved):

PR #398, opened 2025-08-19
Extend the multisource models to take bandflux only models

jeremykubica:
Allow both `AdditiveMultiSourceModel` and `RandomMultiSourceModel` to accept bandflux-only models. Add tests to confirm that this works and that a combination of bandflux and SED models works.

Also add a notebook showing how the combination of models can be used in the microlensing context.


docs/notebooks/microlensing.ipynb
```diff
@@ -323,6 +323,73 @@
     "plt.ylabel(\"Flux\")\n",
```
hombit, line 323: <p>This example feels a little bit artificial: we usually observe μ-lensing in dense fields of Milky Way, so blends are from other field stars, not galaxies. I'd keep the model two component (or hierarchical three): lensed source + (lens + blends), where (lens + blends) term may be represented by a single constant-magnitude tdastro source.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tdastro/pull/398/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

hombit (approved):
Great, thanks

PR #410, opened 2025-08-29
Initial wrapper for Redback models

jeremykubica:
This PR provides an initial proof of concept wrapper for redback models in TDAstro, including a demo notebook.

### DIFF
--- a/docs/notebooks.rst
+++ b/docs/notebooks.rst
@@ -57,6 +57,7 @@ The following notebooks provide example simulations using the TDAstro package.
     Microlensing Effect Example <notebooks/pre_executed/microlensing>
     PLAsTiCC SNIa <notebooks/pre_executed/plasticc_snia>
     PZFlow Source Demo <notebooks/pre_executed/using_pzflow>
+    Redback Models Demo <notebooks/pre_executed/redback_example>
     Resampling LCLIB <notebooks/pre_executed/lclib_example>
     Synphot-based Models <notebooks/pre_executed/synphot_example>
 


src/tdastro/sources/redback_models.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # Create the source itself.
+        if isinstance(source, str):
+            try:
+                import redback
+            except ImportError as err:
+                raise ImportError(
+                    "redback package is not installed be default. To use the RedbackWrapperModel, "
+                    "please install redback. For example, you can install it with "
+                    "`pip install redback` or `conda install conda-forge::redback`."
```
hombit, line 96: It is not on conda yet https://anaconda.org/search?q=redback

```suggestion
                    "`pip install redback`."
```

src/tdastro/sources/redback_models.py
```diff
@@ -0,0 +1,200 @@
...
+            t0 = 0.0
+        shifted_times = times - t0
+
+        # Redback takes the wavelengths in nanometers and produces spectra in
+        # erg / cm^2 / s / Angstrom.  The spectra is oversampled in time, so we
+        # will need to interpolate it at the times of interest.
+        wavelengths_nm = wavelengths / 10.0
+        rb_result = self.source(
+            shifted_times,
+            lambda_array=wavelengths_nm,
```
hombit, line 185: It actually looks like `redback` uses angstroms
https://github.com/nikhil-sarin/redback/blob/33a4a493a78dca8d0cdacdd38f38b2519884e38d/redback/transient_models/kilonova_models.py#L1424
https://github.com/nikhil-sarin/redback/blob/33a4a493a78dca8d0cdacdd38f38b2519884e38d/redback/utils.py#L255

src/tdastro/sources/redback_models.py
```diff
@@ -0,0 +1,200 @@
...
+            output_format="spectra",
```
hombit, line 186: It looks like the models are not vectorized over frequencies, so it is probably right to use "spectra" instead of "flux_density" here, but I'm not really sure about that.

hombit (changes requested):
Do we add redback to the optional deps?

jeremykubica:
I addressed Kostya's comments, but given all the renamining PRs submitted in the past few weeks, it is cleanest to just make a new PR. I will do that shortly.

PR #412, opened 2025-08-29
Create a new ObsTable base class

jeremykubica:
This PR is the first step in refactoring `OpSim` to generalize to more input formats and surveys. This was broken off into its own step and implemented in such a way that it should be fully backwards compatible.

The PR creates a new, more abstract `ObsTable` class that holds the core functionality we want for surveys. A few notable changes to the internal structures:

1) `table` is no longer meant to be accessed directly.

2) `colmap` has been replaced with `inv_colmap` and all the column names are remapped to "standard names". This is the inverse of how it was previously done in `OpSim` where the original names were kept and `colmap` provided a mapping into those. There are a few reasons for this. First it makes the logic within the survey's easier, because everything uses the same column names. Second, it allows us to save and load the data set without worrying about saving/loading the column mapping. Third, it is a step toward allowing us to concatenate surveys (although there are a bunch more changes needed for this).

3) The survey's values (like `dark_current`) are stored as a dictionary instead of object attributes. While this makes them a little more cumbersome to access, it will allow easy saving and loading by putting these in metadata (a future change will be to allow the `ObsTable` to be saved as an astropy table with metadata).


hombit (commented):
It looks good! It looks like a lot of refactoring happening here, so I haven't really read all the changes. Let's discuss the base class name on the meeting today, @mi-dai proposed `ObsTable` which I also like

hombit (approved):

PR #431, opened 2025-09-03
Allow ObsTable to read/write using parquet

jeremykubica:
Create help functions for writing a `ObsTable` to parquet and reading it in from parquet. In addition allow the `ObsTable` to extract survey values from the metadata of a Pandas dataframe.

Also fixes some naming and comments left over from the `OpSim` -> `ObsTable` transition.


hombit (approved):
Thanks, looks great

PR #432, opened 2025-09-03
Add support for processing multiple surveys

### LINKED ISSUE #417: Add ability to simulate from multiple surveys
Extend the `simulate_lightcurves` function to simulate observations from multiple surveys at the same time. This would require extending the function to:

- Take a list of `ObsTable` and `PassbandGroup` (lists must be the same size)
- Generate the parameters for each object once (to ensure all surveys use the same underlying object parameters), then generate the observed bandfluxes under the different surveys.
- Concatenating the observations
- Adding a field to indicate which survey each observation came from

jeremykubica:
Closes #417 

Allow TDAstro to generate observations from multiple surveys by taking a list of `ObsTable` and `PassbandGroup`. Which survey is used is tracked by an additional column in the lightcurve (`survey_idx`).

### DIFF
--- a/src/tdastro/obstable/obs_table.py
+++ b/src/tdastro/obstable/obs_table.py
@@ -17,7 +17,8 @@ class ObsTable:
     Parameters
     ----------
     table : dict or pandas.core.frame.DataFrame
-        The table with all the survey information.
+        The table with all the survey information. Metadata can be included in the
+        "tdastro_survey_data" entry of the attributes dictionary.
     colmap : dict, optional
         A mapping of standard column names to their names in the input table.
         For example, in Rubin's OpSim we might have the column "observationStartMJD"
@@ -53,6 +54,7 @@ class ObsTable:
         "radius": None,
         "read_noise": None,
         "zp_per_sec": None,
+        "survey_name": "Unknown",
     }
 
     def __init__(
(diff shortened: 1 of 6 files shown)


mi-dai (commented):
Looks great! One comment - this is assuming all the surveys will use the same "model", but it may be more likely that for different surveys the parameter distributions will change (even though it's the same underlying SED like SALT), then it becomes a different "model". Do we also want to support a list of models? (Or users can probably just write a loop themselves)

jeremykubica:
In what ways would the distribution of parameter change? Are you thinking that you would see dimmer objects in some surveys or some objects only in infrared? Is that what you mean?

In that case it would be best to loop over the models.

But let me know if you have something else in mind.

jeremykubica:
The simulate function is meant to generate observations from sample objects from a single population. I don't think we should extend it to handle multiple populations. That should be done via an outer loop that iterates over all populations we wish to sample and generates observations from them. The different populations would be different sets of rows in the output table, so we are concatenating results data.

Within the simulate functions, I do think we want functionality for generating all observations for a given object (even if they are observed with different surveys). As opposed to different rows, we want observations of the same object by different surveys to share parameter data and be in the same row.

mi-dai (approved):

PR #436, opened 2025-09-04
Fix name of healpix list

jeremykubica:
The data stored in `ApproximateMOCSampler` is a list of healpix IDs at a given depth instead of an actual MOC. This PR fixes the name.


hombit (approved):
Thank you!

PR #461, opened 2025-09-17
Clarify the usage of the light curve reference epoch

### LINKED ISSUE #458: Bug when setting t0 for MultiLightcurveTemplateModel?
**Bug report**

**Before submitting**
Please check the following:

- [x ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.

Hi, it's me again. I was wondering about the meaning of setting t0 when creating a model for simulations. Just for testing, I'm using a single model, simulated once, setting both the position and t0:

```
source = MultiLightcurveTemplateModel([all_lcs[260]],
    passband_group,
    t0=t0_source,
    ra=ra_source,
    dec=dec_source,
    #filters=filters,
    node_label="source",
)
```
`all_lcs` is a list of LightcurveData objects. t0_source is set to `63426.03186743915`. After simulating 1 instance with `simulate_lightcurves(source, num_samples=1, obstable=opsim, passbands=passband_group)`, inspecting light curves.params[0]['source.t0'] indeed gives me the correct value. However, when plotting the curves, t0 seems to be at the end of the event (red dashed line). For comparison, the perfect curve is in solid black line. Is this intended behavior? I'd have thought more useful for t0 to be the beginning of the event, shifted form the perfect data. Maybe there's something to do with how the light curves are stored, since in the second plot there are negative days. I don't really know.

Thanks and sorry for creating lots of issues in a short period of time.

Image

Image

### LINKED ISSUE #460: Rename `lc_t0` and make it mandatory
**Feature request**

We found that `lc_t0` causes a lot of confusion (see #458 and also from the Tucson hack). I think we should:
1. Make it a mandatory argument, so users are aware that it exists
2. Rename it to e.g. `lc_data_t0`

**Before submitting**
Please check the following:

jeremykubica:
Closes #460 (though not in the way proposed)
Closes #458 

This changes how the reference epoch is passed to and used by the various light curve template models. The reference epoch is used to shift the light curve so the data's time stamp of 0.0 aligns with the models `t0`. In most cases we want the reference epoch to be the time stamp of the first sample in the light curve data. This is when the light curve "starts".

The change automatically zero-shifts the light curve so the reference epoch is the first time stamp. It provides the user the ability to forcibly override this behavior using the `forced_lc_t0` to choose another reference epoch within the light curve. However this requires the user to look through the light curve and understand what point should correspond to `t0`, so we want the default behavior to be that this is handled automatically.

### DIFF
--- a/docs/notebooks/lightcurve_source_demo.ipynb
+++ b/docs/notebooks/lightcurve_source_demo.ipynb
@@ -66,9 +66,11 @@
    "source": [
     "## Creating the model\n",
     "\n",
-    "We create the model from light curves for each passbands of interest. This is what we want the model to reproduce then we call evaluate_bandfluxes(). While these light curves do no need to be the same as in the `PassbandGroup` every light curve must have a corresponding entry in the `PassbandGroup`.\n",
+    "We create the model from light curves for each passbands of interest. This is what we want the model to reproduce then we call `evaluate_bandfluxes()`. While these light curves do no need to be the same as in the `PassbandGroup` every light curve must have a corresponding entry in the `PassbandGroup`.\n",
     "\n",
-    "For simplicity of the demo, we create each curve as a randomly parameterized sin wave.  Note that the times for all the light curves do not need to be the same."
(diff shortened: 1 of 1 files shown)


jeremykubica:
Replaced by #462

PR #474, opened 2025-09-19
Provide the foundation for manual surveys

jeremykubica:
Provide a survey class that takes in basic pointing data specified as a table or dictionary. Instead of using instrument-specific details to generate the noise, we provide a bunch of options for manually specifying noise, including:
  * Constant
  * Constant per-band
  * Given (list or numpy array)
  * A function
 
This class will serve as the basis for randomly generated surveys, such as uniform cadence,  poisson cadence, etc.


jeremykubica:
As discussed in the team meeting, we will create a hybrid noise model. Will reopen as a new PR.

PR #481, opened 2025-09-19
Allow surveys to use custom footprints

jeremykubica:
Create a `SurveyFootprint` class to represent, query, and plot the survey's footprint. We also include two basic footprints for testing (circular and rectangular). Extend `ObsTable` to perform footprint filtering if a footprint was provided.


mi-dai:
Naming suggestion: DetectorFootprint instead of SurveyFootprint

jeremykubica:
> And a question: can this include gaps on the detectors?

In the future "yes". The two simple footprints included with this PR represent simple solid regions only. But we can add more complex footprints, including chip gaps.

jeremykubica:
I found a bug in how this work, so I am going to close this PR while I fix the bug.

PR #485, opened 2025-09-20
Create a more realistic fake survey

jeremykubica:
Provide a survey class that takes in basic pointing data specified as a table or dictionary and uses a combination of: 1) data from the table, 2) constants provided in the constructor, and 3) reasonable defaults to compute zero points and noise.

The class also has an override for testing that allows it to produce constant noise.

### DIFF
--- a/src/lightcurvelynx/obstable/fake_obs_table.py
+++ b/src/lightcurvelynx/obstable/fake_obs_table.py
@@ -9,11 +9,8 @@ class FakeObsTable(ObsTable):
     """A subclass for a (simplified) fake survey. The user must provide a constant
     flux error to use or enough information to compute the poisson_bandflux_std noise model.
     To compute the flux error, the user must provide the following values
-    either in the table or as keyword arguments to the constructor:
-    - fwhm_px : The full-width at half-maximum of the PSFs in pixels.
-    - sky : The sky backgrounds  in the units of electrons / pixel^2.
-    - zp_per_band : Instrumental zero points in nJy
-    Each of these three parameters is specified as a dictionary mapping filter names to values.
+    either in the table or as keyword arguments to the constructor 1) sky, 2) zp_per_band,
+    and 3) either a) fwhm_px or b) footprint.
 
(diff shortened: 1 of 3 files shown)


src/lightcurvelynx/obstable/fake_obs_table.py
```diff
@@ -0,0 +1,169 @@
+import numpy as np
+
+from lightcurvelynx.astro_utils.noise_model import poisson_bandflux_std
+from lightcurvelynx.consts import GAUSS_EFF_AREA2FWHM_SQ
+from lightcurvelynx.obstable.obs_table import ObsTable
+
+
+class FakeObsTable(ObsTable):
+    """A subclass for a (simplified) fake survey. The user must provide the provide a constant
```
mi-dai, line 9:
```suggestion
    """A subclass for a (simplified) fake survey. The user must provide a constant
```

src/lightcurvelynx/obstable/fake_obs_table.py
```diff
@@ -0,0 +1,169 @@
...
+        - read_noise: The read noise for the camera in electrons (default=0.0).
+        - sky_background (if not in table) in the units of electrons / pixel^2 (default=None).
+        - survey_name: The name of the survey (default="FAKE_SURVEY").
+    """
+
+    # Default survey values.
+    _default_survey_values = {
+        "dark_current": 0,
+        "exptime": 30,  # seconds
+        "fwhm": None,  # pixels
```
mi-dai, line 51: psf should be at least per band. We should also allow giving noise equivalent area directly (instead of calculating from fwhm)

src/lightcurvelynx/obstable/fake_obs_table.py
```diff
@@ -0,0 +1,169 @@
...
+        "nexposure": 1,  # exposures
+        "radius": None,  # degrees
+        "read_noise": 0,  # electrons
+        "sky_background": None,  # electrons / pixel^2
```
mi-dai, line 55: sky_background should also be at least per band

PR #498, opened 2025-09-24
Expand the LightCurveTemplate models to take magnitudes

jeremykubica:
This is an initial step to getting Amy's notebook to run through LightCurveLynx. The inputs in that notebook are all light curves defined in terms of magnitude instead of flux.


hombit (approved):
Thank you!

PR #500, opened 2025-09-24
Rename footprint

### LINKED ISSUE #499: Rename `FakeObsTable(footprint)`
Sorry for the late thought: footprint may be confusing for the obstable, because it could mean both "survey footprint", and "PSF footprint". Maybe we should rename to `psf_footprint` or `psf_area`.

jeremykubica:
Closes #499 

Renames the `footprint` parameter in both the ObsTable classes and the noise functions to `psf_footprint` to avoid any confusion with the detector's footprint.

### DIFF
--- a/src/lightcurvelynx/astro_utils/noise_model.py
+++ b/src/lightcurvelynx/astro_utils/noise_model.py
@@ -9,7 +9,7 @@ def poisson_bandflux_std(
     *,
     total_exposure_time: npt.ArrayLike,
     exposure_count: npt.ArrayLike,
-    footprint: npt.ArrayLike,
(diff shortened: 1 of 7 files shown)


src/lightcurvelynx/astro_utils/noise_model.py
```diff
@@ -9,7 +9,7 @@ def poisson_bandflux_std(
     *,
     total_exposure_time: npt.ArrayLike,
     exposure_count: npt.ArrayLike,
-    footprint: npt.ArrayLike,
+    psf_footprint: npt.ArrayLike,
```
hombit, line 12: Here we don't have a confusion, but I think it is fine to also rename it

PR #502, opened 2025-09-24
replace time with obsmjd in obstable calculation to avoid error about…

mi-dai:
`time` column was assigned in ZTFObsTable (when converting `obsdate` to `obsmjd`, which will cause error when using a new colmap that include `time`. Changed to `obsmjd`.


jeremykubica:
This is going to give an error during the tests because there is already a mapping in `_default_colnames` doing this mapping. 

What is the underlying problem? Are you trying to read a ZTF table that has a time column (with values you don't want)?

jeremykubica (approved):

PR #503, opened 2025-09-25
Move ZTF zeropoint code into zeropoint.py

jeremykubica:
Small reorganization PR. Move the code to compute the zero points from maglim into the zeropoint.py file. Call it from ZTFObsTable with the ZTF default parameters.


mi-dai (approved):
Looks good. Although the units in the function is somehow tailored to ZTF and not general enough. I would merge this now and generalize it later.

PR #507, opened 2025-09-25
Transform a light curve into a redback Transient

jeremykubica:
Convert the light curve data into mangnitudes and use it to create a redback Transient object.


jeremykubica:
Closing as we are adding the interface on the redback side: https://github.com/nikhil-sarin/redback/pull/293

PR #513, opened 2025-09-26
Remove rest_frame parameter from some effects

jeremykubica:
To avoid confusion remove the `rest_frame` parameter from the `ConstantDimming` and `Microlensing` effects. Set the default in the Microlensing case to observer frame.


hombit (approved):
Great!

PR #516, opened 2025-09-29
Allow multi-processing of results via a pool parameter

jeremykubica:
Allow the core simulation function to take an executor parameter and distribute the tasks via that executor. This allows the simulation to be parallelized by a variety of mechanisms including ProcessPoolExecutor, Dask, and Ray. 

This PR also provides the option (via argument) to output the results to a file instead of returning the NestedFrame so that we can distribute computation without worrying about all of the results fitting in memory together.

### DIFF
--- a/src/lightcurvelynx/simulate.py
+++ b/src/lightcurvelynx/simulate.py
@@ -7,6 +7,7 @@ from nested_pandas import NestedFrame
 
 from lightcurvelynx.astro_utils.noise_model import apply_noise
 from lightcurvelynx.models.physical_model import BandfluxModel
+from lightcurvelynx.utils.post_process_results import concat_results
 
 
 class SimulationInfo:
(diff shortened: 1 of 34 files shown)


src/lightcurvelynx/simulate.py
```diff
@@ -221,10 +333,100 @@ def simulate_lightcurves(
...
+        the results (instead of just the full dictionary of parameters). These
+        must be specified as strings in the node_name.param_name format.
+        If None, no additional columns are saved.
+    rng : numpy.random._generator.Generator, optional
+        A given numpy random number generator to use for this computation. If not
+        provided, the function uses the node's random number generator.
+    generate_citations : bool, optional
+        If True, generate citations for the simulation and output them.
+        Default is False.
+    pool : multiprocessing.Pool, optional
```
hombit, line 395: `multiprocessing.Pool` is considered an older implementation, while `concurrent.futures.Executor` is a base class for a more modern Python standard library parallelism implementation. See the notes in the `multiprocessing` docs:
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.ThreadPool

src/lightcurvelynx/simulate.py
```diff
@@ -221,10 +333,100 @@ def simulate_lightcurves(
...
+        obstable=obstable,
+        passbands=passbands,
+        rng=rng,
+        apply_obs_mask=apply_obs_mask,
+        time_window_offset=time_window_offset,
+        obstable_save_cols=obstable_save_cols,
+        param_cols=param_cols,
+    )
+
+    if pool is not None and batch_size < num_samples:
```
hombit, line 419: I'd always use the provided pool, especially if it represents some distributed cluster.

hombit (commented):
I'm not fully convinced that we should have this code at Lynx. First, we make a lot of decisions here about the data flow control, e.g. in this implementation the whole obstable would be pickled and transferred, the whole result dataframe will be pickled and transferred, and all the results (doubled) would be in memory at a few points.

I see different pipeline control decisions which potentially could be made:  
1. Pre-generate parameters, pre-select obstable values.  
2. Do not transfer the result back, just dump them as a parquet file.  
3. `pool.map` is a common interface, but others exist. `.submit/.result` interface would allow a progress bar!

src/lightcurvelynx/simulate.py
```diff
@@ -221,20 +354,157 @@ def simulate_lightcurves(
...
+        the results (instead of just the full dictionary of parameters). These
+        must be specified as strings in the node_name.param_name format.
+        If None, no additional columns are saved.
+    output_file_path : str or Path, optional
+        The file path and name of where to save the results. If provided the results
+        are saved to this file instead of being returned directly.
+    rng : numpy.random._generator.Generator, optional
+        A given numpy random number generator to use for this computation. If not
+        provided, the function uses the node's random number generator.
+    executor : concurrent.futures.Executor, optional
```
hombit, line 456: Let's maybe put a little more documentation about what we expect to get here? Because we are kinda supporting duck-typed executors and futures here.

hombit (approved):

PR #526, opened 2025-09-30
add zero point uncertainty to error calculation

mi-dai:
Add zero point error. I'm keeping it as a constant value right now.


hombit (commented):
I checked the equation, it is correct!

Let's also add it to the base class, and inherit it from there, making it a constructor parameter for all obstables?

mi-dai:
base class doesn't have a noise implementation right now. https://github.com/lincc-frameworks/lightcurvelynx/blob/e396742fee2dbff2f1ce44f56096b1f5bcf4758b/src/lightcurvelynx/obstable/obs_table.py#L675

hombit (approved):

PR #531, opened 2025-10-01
Add checks on how redshift is set in sncosmo models

jeremykubica:
In my debugging notebook I was setting redshift through the `z` parameter for both the sncosmo model (correct way to do it) and the lightcurvelynx model (incorrect way). This PR checks for conflicts between an incoming `z` parameter and a `redshift` parameter.


jeremykubica:
You are completely correct. The sncosmo source does not take z. I made an error in the analysis notebook where I was setting "z" for use in the sncosmo model, but passing the same params into the SncosmoWrapperModel (instead of passing in redshift). Maybe this is an overly specific case because I was trying to work with two different packages at the same time.  What do you think?

jeremykubica:
Closing after discussion with Mi.

PR #538, opened 2025-10-02
Change use of filters to obs index in saturation computation

### LINKED ISSUE #535: Use index instead of filters in compute_saturation
To make it more flexible for a future transition to electron/ADU-based thresholding, let's get index instead of filters (just like in `bandflux_error_point_source` method), which would later used for PSF, etc

_Originally posted by @hombit in https://github.com/lincc-frameworks/lightcurvelynx/pull/521#discussion_r2392486257_

olivialynn:
Closes #535 
This ended up being way shorter than I expected - I guess I just needed to look at it with a fresh set of eyes : )


hombit (approved):
Thanks!

PR #545, opened 2025-10-08
Allow sampling RA, dec from a HATS catalog

jeremykubica:
This PR extends the `ObsTableRADECSampler` to take data from more sources, including loading in a HATS catalog. 

Also fixes the sampling, using the approach provided by @hombit , to be uniform around the points.

### DIFF
--- a/src/lightcurvelynx/math_nodes/ra_dec_sampler.py
+++ b/src/lightcurvelynx/math_nodes/ra_dec_sampler.py
@@ -127,6 +127,11 @@ class ObsTableRADECSampler(TableSampler):
     def from_hats(cls, path, *, radius=None, extra_cols=None, in_order=False, **kwargs):
         """Create a GivenRADECSampler from the observations in a HATS Catalog.
 
+        Note
+        ----
+        If you have an existing Dask client, it may be used.
+        See LSDB documentation for details: https://docs.lsdb.io/en/latest/
+
         Parameters
         ----------
         path : str or Path


src/lightcurvelynx/math_nodes/ra_dec_sampler.py
```diff
@@ -72,42 +74,100 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+        GivenRADECSampler
+            The created GivenRADECSampler object.
+        """
+        # See if the (optional) LSDB package is installed.
+        try:
+            from lsdb import read_hats
+        except ImportError as err:
+            raise ImportError(
+                "The lsdb package is required to read HATS catalogs. "
+                "Please install it via 'pip install lsdb'."
```
hombit, line 159:
```suggestion
                "Please install it via 'pip install lsdb' or 'conda install conda-forge::lsdb'."
```

src/lightcurvelynx/math_nodes/ra_dec_sampler.py
```diff
@@ -72,42 +74,100 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+            ) from err
+
+        # Compute the full list of columns to load from HATS.
+        cols_to_load = ["ra", "dec"]
+        if extra_cols is not None:
+            cols_to_load.extend(extra_cols)
+        columns = list(set(cols_to_load))  # Remove any duplicates.
```
hombit, line 166: I don't know if it matters for the rest of the pipeline, but the order of columns could differ from time to time.

src/lightcurvelynx/math_nodes/ra_dec_sampler.py
```diff
@@ -72,42 +74,100 @@ def compute(self, graph_state, rng_info=None, **kwargs):
...
+        # Add any extra columns (without duplicates).
+        if extra_cols is not None:
+            for col in extra_cols:
+                if col not in data_dict:
+                    data_dict[col] = data[col]
+
         super().__init__(data_dict, in_order=in_order, **kwargs)
 
+    @classmethod
+    def from_hats(cls, path, *, radius=None, extra_cols=None, in_order=False, **kwargs):
```
hombit, line 127: Should we note anything about dask client here?

src/lightcurvelynx/math_nodes/ra_dec_sampler.py
jeremykubica: It shouldn't impact the rest of the pipeline. I was thinking about a few different ways of trying to get around the different names, but don't have a great suggestion beyond people including it in the extra columns.

PR #557, opened 2025-10-14
Refactor Zeropoint Code

### LINKED ISSUE #506: Revisit units in calculate_zp_from_maglim()
Revisit units in calculate_zp_from_maglim() to make it more general/consistent

jeremykubica:
Refactor the zeropoint code with a step toward generalization, including:
- `magnitude_electron_zeropoint` and `flux_electron_zeropoint` do not fall back to LSST defaults. They require explicit values to be passed in.
- `magnitude_electron_zeropoint` and `flux_electron_zeropoint` can take arrays of values instead of just dictionaries (in case the values are in the tables).
- `calculate_zp_from_maglim` uses more standard units (sky background in electrons / pixel).
- Move LSST default values into `OpSim` with the rest of the LSST default values.

Closes #506

### DIFF
--- a/src/lightcurvelynx/astro_utils/zeropoint.py
+++ b/src/lightcurvelynx/astro_utils/zeropoint.py
@@ -6,20 +6,20 @@ import numpy.typing as npt
 from lightcurvelynx.astro_utils.mag_flux import mag2flux
 
 
-def sky_bg_adu_to_electrons(sky_bg_adu: npt.ArrayLike, gain: float | npt.ArrayLike) -> npt.ArrayLike:
+def sky_bg_adu_to_electrons(sky_bg_adu, gain):
     """Convert sky background from ADU/pixel to electrons/pixel.
 
     Parameters
     ----------
-    sky_bg_adu : ndarray of float
+    sky_bg_adu : float or ndarray of float
         Sky background in ADU/pixel.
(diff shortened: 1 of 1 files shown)


src/lightcurvelynx/astro_utils/zeropoint.py
```diff
@@ -5,59 +5,34 @@
...
-Values are from
-https://community.lsst.org/t/release-of-v3-4-simulations/8548/12
-Calculated with syseng_throughputs v1.9
-"""
-
-
-# Suppress "no docstring", because we define it via an attribute.
-def magnitude_electron_zeropoint(  # noqa: D103
+
+def sky_bg_adu_to_electrons(sky_bg_adu: npt.ArrayLike, gain: float | npt.ArrayLike) -> npt.ArrayLike:
```
hombit: I'm not sure about the type hint here. I believe that if it is `ArrayLike`, the function is supposed to use `np.asarray` or `np.asanyarray` to convert it from inputs to something you can use the `*` operator on. Also, if we do this, the output type would be `NDArray` or something similar.

jeremykubica: I removed the type hints and expanded the documentation to make it clearer that this can operate on more types.

PR #559, opened 2025-10-16
Fix broken tests

jeremykubica:
The smoke test was failing because two of the test pzflow model files were using a deprecated format. This PR re-saves them using current libraries and Python3.13.


jeremykubica:
I'm not sure what you mean. The data files were just created by using pzflow's save functionality. I read the old files in with pzflow and resaved them (in python3.10 with the most recent pzflow library) and everything works.

PR #562, opened 2025-10-21
Create a flexible parameter computation for the FakeObsTable 

jeremykubica:
Create an extendable `_ParamDeriver` class that tries to derive missing parameters for the a `ObsTable` using a series of rules. This class is used in `FakeObsTable` to provide flexibility and compute zeropoints from what it can find.


hombit (approved):
Let's remove LSST-related stuff such as airmass, ext_coeff, etc for now?

PR #570, opened 2025-10-29
Support reading and Querying of SIMSED models

jeremykubica:
While the ultimate goal of this PR is to support reading in an querying SIMSED models, we use it as an opportunity to create a more general SED-based lightcurve class. This class will eventually replace `SplineModel` and have additional readers to incorporate data from other systems.  For now we just mirror the functionality of the `LightcurveTemplateModel` class and add readers for SIMSED data.

### DIFF
--- a/tests/lightcurvelynx/models/test_sed_template_model.py
+++ b/tests/lightcurvelynx/models/test_sed_template_model.py
@@ -242,6 +242,6 @@ def test_slsn_simsed_model(test_data_dir) -> None:
     assert sed_values.shape == (2, 1)
     assert np.all(sed_values > 0.0)
 
-    # We know that the expected magntiudes for a SLSN at 10 Mpc will be around -22.
+    # We know that the expected magntiudes for a SLSN at 10 pc will be around -22.
     mag_vals = flux2mag(sed_values.flatten())
     assert np.all((mag_vals > -23.0) & (mag_vals < -21.0))


tests/lightcurvelynx/data/fake_simsed/SED.INFO
hombit: Please add some comments here, or into a README file in the same directory explaining where this data came from

src/lightcurvelynx/models/sed_curve_model.py
```diff
@@ -0,0 +1,520 @@
...
+            A dictionary of parameters read from the SED.INFO file.
+        """
+        info_file = simsed_dir / "SED.INFO"
+        if not info_file.exists():
+            raise FileNotFoundError(f"SED.INFO file not found in {simsed_dir}.")
+        logger.debug(f"Reading SIMSED data from {info_file}.")
+
+        parameters = {}
+        lc_files = []
+        with open(info_file, "r") as f:
```
hombit: The header is promised to be YAML-compatible, maybe it can make parsing easier.

src/lightcurvelynx/models/sed_curve_model.py
```diff
@@ -0,0 +1,520 @@
...
+            scaled by the light curve values.
+        """
+        # Use the light curve selected by the sampler node to compute the flux density.
+        model_ind = self.get_param(graph_state, "selected_lightcurve")
+        shifted_times = times - self.get_param(graph_state, "t0")
+        return self.lightcurves[model_ind].evaluate_sed(shifted_times, wavelengths)
+
+
+class SIMSEDModel(MultiSEDCurveModel):
+    """Generate lightcurves from SIMSED-formated data.
```
hombit: Let's say here that the format is specified by SNANA

src/lightcurvelynx/models/sed_curve_model.py
```diff
@@ -0,0 +1,520 @@
...
+from scipy.interpolate import RectBivariateSpline
+
+from lightcurvelynx.astro_utils.unit_utils import flam_to_fnu
+from lightcurvelynx.math_nodes.given_sampler import GivenValueSampler
+from lightcurvelynx.models.physical_model import SEDModel
+
+logger = logging.getLogger(__name__)
+
+
+class LightcurveSEDData:
```
hombit: Sorry for the naming concern, but let's maybe refer to it without "lightcurve" term, which is usually applied to bandflux/mag data. Maybe TemporalSEDData or just SEDData? The same goes to lightcurves argument, etc

src/lightcurvelynx/models/sed_curve_model.py
```diff
@@ -0,0 +1,520 @@
...
+        lightcurve_matrix = np.zeros((len(unique_phases), len(unique_wavelengths)))
+        for wave, phase, flux_val in zip(wavelengths, phases, fluxes, strict=False):
+            wave_idx = np.where(unique_wavelengths == wave)[0][0]
+            phase_idx = np.where(unique_phases == phase)[0][0]
+            lightcurve_matrix[phase_idx, wave_idx] = flux_val
+
+        return unique_phases, unique_wavelengths, lightcurve_matrix
+
+
+class SEDCurveModel(SEDModel):
```
hombit: Again, let's try to avoid lightcurves in the context of SED models (until we talk about generated bandfluxes)

src/lightcurvelynx/models/sed_curve_model.py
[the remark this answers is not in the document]
jeremykubica: It is the SED version of the `LightcurveBandData` ([here](https://github.com/lincc-frameworks/lightcurvelynx/blob/6da7ad8316d43b6e74f556f2799165b39bc36648/src/lightcurvelynx/models/lightcurve_template_model.py#L27)). It wraps a grid of SED values and its interpolator. It also stores and processes meta data for the lightcurve, such as the t0 of the data (`lc_data_t0`), whether the curve is periodic, etc.

The class effectively represents a interpolation based model of raw SED data. The reason we break these functions into a separate class (instead of putting them in the model class) is so that we can store multiple light curves in the `MultiSEDCurveModel` class without replicating the parameters.

hombit (approved):
Thank you! My main concerns are about using (light)curve naming.

A demo with SNANA data would be useful, let's make an issue for that!

PR #573, opened 2025-10-29
Create SNANA demo notebook

### LINKED ISSUE #563: Support SIMSED input files
**Feature request**

Supporting the same input SED files as SNANA would give us access to a lot of pre-sampled temporal-spectral models created for PLAsTiCC and ELAsTiCC. Examples of these files may be found in this Zenodo repository, which we can also use to prepare documentation on how to use these inputs: https://zenodo.org/records/2612896. The SNANA manual, section 9.6, has a good overview of the models, but I'll try to summarize so-called SIMSED models here.

**Directory structure**
Each model is represented by an "SED.INFO" file with some metadata and a list of tabular data files. Each table represents a single "physical source" for some parameters pre-sampled and listed in the metadata file.

**SED tables**
Each table has three columns: phase in days, wavelength in angstroms, and "flux." "Flux" is a bit unusual: it is in erg/cm²/s/Å as seen from 10 pc in rest-frame (see also FLUX_SCALE below), which can be converted to rest-frame units we use by something like:
```python
flam_to_fnu(
    flux_flam = flux_from_table * flux_scale * (10 / luminosity_distance_pc)**2,
    wavelengths = waves_from_table,
    wave_unit=u.AA,
    flam_unit=u.erg / u.s / u.cm**2 / u.AA,
    fnu_unit=u.mJy,
)
```

There are two formats for these tables: textual, which is just `\s+` separated columns with no headers (see e.g., "SIMSED.SLSN-I-MOSFIT.tar.gz"), and binary, which is used to speed up textual file reads. I don't think binary SIMSED files are being distributed, so we probably don't have to support them. We still may benefit from creating this kind of files ourselves on the first read and memory-mapping them, so we do not explode memory usage.

**SED.INFO**

The metadata file has two sections: header and a table of the SED files. The header has some fields we can ignore (or ignore for now), such as mag error, etc. We do need `FLUX_SCALE`, which is the value the flux column should be multiplied by, and we may use `RESTLAMBDA_RANGE`, which specifies the minimum and maximum wavelength values in angstroms.

The table below the header is straightforward: it gives the filename and parameter values used, which we should just move around.

**Before submitting**
Please check the following:

jeremykubica:
Closes #563


docs/notebooks/pre_executed/snana_example.ipynb
```diff
@@ -0,0 +1,298 @@
+{
```
hombit: <p>Let's put some redshift distribution in? So we have realistic fluxes. We use uniform between 0.1 and 0.6 in another notebook, it is not realistic, but still better</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lightcurvelynx/pull/573/discussion/#comment-2474972818'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

hombit: <p>Something weird happened here...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lightcurvelynx/pull/573/discussion/#comment-2474972828'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

mi-dai: <p>zp in ObsTable above seems too small?</p>

jeremykubica: I added more to the notebook with a more realistic opsim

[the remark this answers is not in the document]
hombit: <p>Sorry, I meant that the output of one of the previous cells is 699 (which is the right number), but you wrote 700 here.</p>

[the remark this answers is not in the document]
hombit: <p>It says "around", sorry</p>

hombit (approved):

PR #589, opened 2025-11-10
Rename ConstantDimming to ScaleFluxEffect

jeremykubica:
Rename the `ConstantDimming` effect to `ScaleFluxEffect` to reflect the fact it can be used with scales > 1.0. Also add text to the notebooks about how this can be used to change the amplitude of given curves and SEDs.

### DIFF
--- a/tests/lightcurvelynx/effects/test_basic_effects.py
+++ b/tests/lightcurvelynx/effects/test_basic_effects.py
@@ -33,7 +33,7 @@ def test_scale_flux_effect_bandflux() -> None:
     with pytest.raises(ValueError):
         _ = effect.apply_bandflux(values)
 
-    # We fail is we use a negative flux scale.
+    # We fail if we use a negative flux scale.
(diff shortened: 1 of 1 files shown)


tests/lightcurvelynx/effects/test_basic_effects.py
```diff
@@ -1,54 +1,58 @@
...
-    values = effect.apply_bandflux(values, flux_fraction=0.1)
+    effect = ScaleFluxEffect(flux_scale=0.1)
+    values = effect.apply_bandflux(values, flux_scale=0.1)
     assert np.all(values == 10.0)
 
     # We fail if we do not pass in the expected parameters.
     with pytest.raises(ValueError):
         _ = effect.apply_bandflux(values)
 
+    # We fail is we use a negative flux scale.
```
olivialynn:
```suggestion
    # We fail if we use a negative flux scale.
```

PR #591, opened 2025-11-11
Expand the documentation

jeremykubica:
Add a new documentation page on the format of the results and how to work with them.


docs/notebooks/introduction.ipynb
```diff
@@ -53,9 +53,7 @@
...
-    "\n",
-    "LightCurveLynx provides tools for generating parameters from a range of models and distributions. For example we can sample (RA, dec) directly from the survey data itself. For more information on how to define the parameter settings, see the `sampling.ipynb` notebook."
+    "We can demonstrate this simulation flow using `SinWaveModel`, a toy model that generates fluxes using a sin wave. The `SinWaveModel` object uses multiple parameters to generate its flux, so we need to speccify how to set these. For some parameters we may have a fixed value, such as a brightness of 100.0. But in most simulations we will want to values of the parameters themselves to vary. We can set these from other nodes (any object that generates or uses parameters). Below we set two of the model's parameters (`frequency` and `t0`) from uniform distributions and two (`RA` and `dec`) are chosen from a Gaussian that matches the toy survey information we will load later in this notebook."
```
olivialynn:
```suggestion
    "We can demonstrate this simulation flow using `SinWaveModel`, a toy model that generates fluxes using a sin wave. The `SinWaveModel` object uses multiple parameters to generate its flux, so we need to specify how to set these. For some parameters we may have a fixed value, such as a brightness of 100.0. But in most simulations we will want the values of the parameters themselves to vary. We can set these from other nodes (any object that generates or uses parameters). Below we set two of the model's parameters (`frequency` and `t0`) from uniform distributions and two (`RA` and `dec`) are chosen from a Gaussian that matches the toy survey information we will load later in this notebook."
```

docs/results_and_output.rst
```diff
@@ -0,0 +1,105 @@
...
+Pandas DataFrame with one row for each observation.
+
+.. code-block:: python
+
+    lightcurve = results["lightcurve"].iloc[0]
+    print("The first object's light curve:")
+    print(lightcurve)
+
+
+Each row also contains a copy of the parameters use to simulate that object (in the ``params`` column), allowing the
```
olivialynn:
```suggestion
Each row also contains a copy of the parameters used to simulate that object (in the ``params`` column), allowing the
```

olivialynn (approved):
LGTM!

PR #592, opened 2025-11-11
Add debugging tools and notebook

jeremykubica:
Add some visualization and graph query tools to help debug models. Also adds a notebook showing how the user can dig into a model.

### DIFF
--- a/docs/notebooks/debugging.ipynb
+++ b/docs/notebooks/debugging.ipynb
@@ -66,7 +66,7 @@
    "id": "02f06d10",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


docs/notebooks/debugging.ipynb
```diff
@@ -0,0 +1,234 @@
...
+    "state = model.sample_parameters()\n",
+    "print(state)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "02f06d10",
+   "metadata": {},
+   "source": [
+    "Our model consists a four nodes. The top-level nodes uses numpy to generate a random brightness for the object (uniformly [1000.0, 3000.0]) and a background (uniformly [100.0, 500.0]. The second level node simply adds the two nodes together to produce a `total_brightness` function. The final node creates a constant SED model with the given brightness. Thus the realized brightness of the sampled object will depend on a chain of nodes and two randomly sampled parameters."
```
hombit:
```suggestion
    "Our model consists of four nodes. The top-level nodes uses numpy to generate a random brightness for the object (uniformly [1000.0, 3000.0]) and a background (uniformly [100.0, 500.0]. The second level node simply adds the two nodes together to produce a `total_brightness` function. The final node creates a constant SED model with the given brightness. Thus the realized brightness of the sampled object will depend on a chain of nodes and two randomly sampled parameters."
```

hombit (approved):

PR #593, opened 2025-11-11
Increase test coverage

jeremykubica:
Increase the code coverage through a combination of:
- adding more tests for edge cases, 
- adding plotting tests to confirm the code executes
- Add a test for `PassbandGroup.process_transmission_tables`
- Add a few "pragma: no cover" annotations (mostly for import errors or errors that should not occur).

### DIFF
--- a/src/lightcurvelynx/astro_utils/passbands.py
+++ b/src/lightcurvelynx/astro_utils/passbands.py
@@ -583,6 +583,30 @@ class PassbandGroup:
         filter_name_mask = np.isin(filters, list(self._filter_to_name.keys()))
         return full_name_mask | filter_name_mask
 
+    def process_transmission_tables(
+        self, *, delta_wave: float | None = 5.0, trim_quantile: float | None = 1e-3
+    ):
+        """Process the transmission tables for all passbands in the group; recalculate group's wave
+        attribute. This function is used to change the preprocessing of the transmission tables after
+        the PassbandGroup has been initialized, such as using different delta_wave or trim_quantile
+        values.
+
+        Parameters
+        ----------
+        delta_wave : float or None, optional
+            The grid step of the wave grid. Default is 5.0.
+        trim_quantile : float or None, optional
+            The quantile to trim the transmission table by. For example, if trim_quantile is 1e-3, the
+            transmission table will be trimmed to include only the central 99.8% of rows.
+        """
+        for passband in self.passbands.values():
+            passband.process_transmission_table(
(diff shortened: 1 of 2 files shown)


olivialynn:
I wonder at the removal of `process_transmission_tables`

At its core, it serves as a convenient method to update the PBG attributes of delta_wave and/or trim_quantile and automatically call `_update_internal_data` afterwards to update the `waves` attribute in the group, which (as the "union of all wavelengths in the passbands") is later used in methods such as:
- `fluxes_to_bandfluxes` to enforce the model fluxes match the grid as the group's waves
- `fluxes_to_bandflux` (via the group's `_in_band_wave_indices`), to construct an index mask of the given flux_density_matrix to focus only on the target passband (and then pass this to the passband's `fluxes_to_bandflux`)

Suggest either:
1. **Updating the docstring** to better explain the intention of `process_transmission_tables` (and/or name change? would reprocess or recalculate be more descriptive?)
2. **Remove the function**
	1. We will trust users who wish to change their group's `delta_wave` or `trim_quantile` values to either initialize a new group, or call `PBG._update_internal_data()` after updating the value via `PBG.delta_wave = 0.1` or so.
		1. in this case, we will want to update the last section of [passband-demo.ipynb](https://github.com/lincc-frameworks/lightcurvelynx/blob/bf5f03607fa68d9d1f86af245dc5ec25541b5d16/docs/notebooks/passband-demo.ipynb) that describes the use (and removes the call to `process_transmission_tables`)
	2. Additionally, we are trusting users who are using PBGs to not make passband-level changes to their transmission tables without manually calling `PBG._update_internal_data()` after. 
		1. This was always a bit of an oversight, I guess, though one might _hope_ it's intuitive not to want your passband grids to diverge if you're using a group ¯\\\_(ツ)\_/¯ . Should probably be addressed regardless - at least a line in the docstring?
		2. I'm worried if we keep `Passband.process_transmission_table()` around, it might imply to users that they need to manually call the method for each passband in a group if they want to change their `delta_wave`/`trim_quantile`. This is both tedious and bad, because it won't trigger the group-based processing described above.
		3. Removing the option to change one's  `delta_wave`/`trim_quantile` on a group level but preserving it on a passband level concerns me a little, then.
		4. Is it enough to add a warning to the docstring? I personally don't trust users to read this.
		5. Do we remove this functionality from the passband-based method as well? Seems like a bit of a refactor.
		6. ??? A third option that I can't think of. I'm going to go ahead and post this now, though, since we're about to meet.

PR #596, opened 2025-11-12
Provide extrapolation options for time

jeremykubica:
For models which have a specified minimum and maximum phase, provide the same extrapolation options as we have for wavelength (zero, constant, linear, exponential). Adds more tests, including checking that we can interpolate on the combination of wavelength and time.

Also adds the ability to use different extrapolators on each side of both wavelength and time.

### DIFF
--- a/src/lightcurvelynx/models/physical_model.py
+++ b/src/lightcurvelynx/models/physical_model.py
@@ -459,7 +459,8 @@ class SEDModel(BasePhysicalModel):
                     "extrapolation using the 'wave_extrapolation' parameter."
                 )
             else:
-                # Add the boundary point at the start for extrapolation and the list to extrapolate.
+                # Add the boundary point at the start for extrapolation and compute
+                # the list of wavelengths to extrapolate.
                 valid_mask = query_waves >= min_valid_wave
                 before_wave_queries = query_waves[~valid_mask]
                 query_waves = np.concatenate(([min_valid_wave], query_waves[valid_mask]))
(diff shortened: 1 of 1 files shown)


src/lightcurvelynx/models/physical_model.py
```diff
@@ -371,51 +414,102 @@ def compute_sed_with_extrapolation(self, times, wavelengths, graph_state, **kwar
...
+                # Compute the flux values after the model's last valid time.
+                flux_density[after_time_mask, :] = self.time_extrapolation.extrapolate_time(
+                    max_valid_time,
+                    flux_density_all_waves[-1, :],
+                    times[after_time_mask],
+                )
+            else:
+                warnings.warn(
+                    "The time_extrapolation is None but some times are outside the model's valid range. "
+                    "These values will be set to zero."
```
hombit: Some more debugging information may be useful. Maybe times you are talking about, or even model parameters?

jeremykubica: I was originally thinking this, but the problem with debugging information is that I think it would issue a unique warning for each set of output instead of a single one.

src/lightcurvelynx/models/physical_model.py
```diff
@@ -249,22 +284,29 @@ class SEDModel(BasePhysicalModel):
         A list of effects to apply in the rest frame.
     obs_frame_effects : list of EffectModel
         A list of effects to apply in the observer frame.
-    wave_extrapolation : WaveExtrapolationModel, optional
+    wave_extrapolation : FluxExtrapolationModel, optional
         The extrapolation model to use for wavelengths that fall outside
         the model's defined bounds.  If None then the model will use all zeros.
+    time_extrapolation : FluxExtrapolationModel, optional
+        The extrapolation model to use for times that fall outside
+        the model's defined bounds.  If None then the model will use all zeros.
```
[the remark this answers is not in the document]
jeremykubica: Done

PR #599, opened 2025-11-17
Add extrapolation for bandflux models

### LINKED ISSUE #598: Add time extrapolation for bandflux models
Extend the extrapolation approaches to handl bandflux only models.

jeremykubica:
Closes #598

This PR adds the ability extrapolate bandflux models over time. Each band is extrapolated based on its own last time and value.

Changes include:
- Adding time extrapolation into bandflux models
- Issuing a warning if `wave_extrapolation` is set for bandflux models since it cannot be used.
- A slight code reorganization to keep things clean where the `compute_bandflux()` is called per filter. This allows us to more easily insert `compute_bandflux_with_extrapolation()` before it.
- Make `minwave`, `minphase`, etc. take kwargs so we can pass extra information (such as filter).

The LightcurveTemplateModel uses the interpolation only if the model is NOT periodic and there is no provided baseline. This keeps the behavior consistent with both the current behavior and the LCLIB files (which have template values).

### DIFF
--- a/src/lightcurvelynx/models/physical_model.py
+++ b/src/lightcurvelynx/models/physical_model.py
@@ -926,7 +926,6 @@ class BandfluxModel(BasePhysicalModel, ABC):
             min_valid_time = min_query_time
         else:
             min_valid_time = min_valid_phase + t0
-        before_time_mask = times < min_valid_time
 
         before_time_queries = None
         if min_query_time < min_valid_time:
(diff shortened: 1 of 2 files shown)


src/lightcurvelynx/models/lightcurve_template_model.py
```diff
@@ -591,6 +598,66 @@ def __init__(
         self.filters = self.lightcurves.filters
         super().__init__(passbands=passbands, filters=self.filters, **kwargs)
 
+        # Raise a warning if time extrapolation is provided but cannot be used.
+        if "time_extrapolation" in kwargs and kwargs["time_extrapolation"] is not None:
+            if periodic:
+                logger.warning("time_extrapolation is provided, but is not used for periodic light curves. ")
```
hombit: Should it be an exception?

PR #620, opened 2025-11-25
Create a bagle wrapper model

jeremykubica:
Create a bagle wrapper model and a notebook that explains how this was done.

PR #622, opened 2025-11-25
Create a demo notebook for wrapping redback models

jeremykubica:
This notebook is part of the same series started in #620, demonstrating how users can wrap their own packages for use in LightCurveLynx

### DIFF
--- a/docs/notebooks/pre_executed/wrapping_redback.ipynb
+++ b/docs/notebooks/pre_executed/wrapping_redback.ipynb
@@ -268,7 +268,7 @@
     "\n",
     "### The Compute Function\n",
     "\n",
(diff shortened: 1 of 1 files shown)


docs/notebooks/pre_executed/wrapping_redback.ipynb
```diff
@@ -0,0 +1,478 @@
+{
```
olivialynn: <pre class="ql-syntax" spellcheck="false">Line #4.    print(sig)  # (time, mass_ejecta, velocity_ejecta, opacity
</pre><p>Should we have a redshift term between time and mass_ejecta?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lightcurvelynx/pull/622/discussion/#comment-2561022641'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

jeremykubica: Yes. I just removed the comment though as it isn't adding anything

PR #626, opened 2025-11-26
Fix a bug in extract_single_sample

jeremykubica:
Make a quick fix to the `copy` and `extract_single_sample` function in `GraphState`. These were not copying over the fixed_var and the sample_offset information. This isn't used anywhere (yet), so it hasn't been causing a problem. But was causing a failure for some of the bagle tests.

### DIFF
--- a/src/lightcurvelynx/graph_state.py
+++ b/src/lightcurvelynx/graph_state.py
@@ -24,6 +24,8 @@ We could have one mean for an object's brightness and another for its positional
 of a host galaxy.
 """
 
+import copy
+
 import numpy as np
 from astropy.io import ascii
 from astropy.table import Table
@@ -195,8 +197,11 @@ class GraphState:
                     new_state.states[node_name][var_name] = var_value
                 else:
                     new_state.states[node_name][var_name] = var_value.copy()
-        for node_name, var_list in self.fixed_vars.items():
-            new_state.fixed_vars[node_name] = var_list.copy()
+
+        # Copy over the sets of fixed variables. This is a single set of strings
+        # per node, so we just need to copy the sets.
+        for node_name, var_set in self.fixed_vars.items():
+            new_state.fixed_vars[node_name] = copy.deepcopy(var_set)
         return new_state
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


src/lightcurvelynx/graph_state.py
```diff
@@ -187,15 +187,16 @@ def copy(self):
...
             new_state.states[node_name] = {}
             for var_name, var_value in node_vars.items():
                 if self.num_samples == 1:
                     new_state.states[node_name][var_name] = var_value
                 else:
                     new_state.states[node_name][var_name] = var_value.copy()
-        for node_name, fixed_vars in self.fixed_vars.items():
-            new_state.fixed_vars[node_name] = fixed_vars.copy()
+        for node_name, var_list in self.fixed_vars.items():
+            new_state.fixed_vars[node_name] = var_list.copy()
```
hombit: What are `var_list` elements? Should it be `copy.deepcopy` instead?

src/lightcurvelynx/graph_state.py
```diff
@@ -502,13 +503,19 @@ def extract_single_sample(self, sample_num):
         # Make a copy of the GraphState with exactly one sample.
         new_state = GraphState(1)
         new_state.num_parameters = self.num_parameters
+        new_state.sample_offset = self.sample_offset
```
hombit: Is it related to https://github.com/lincc-frameworks/lightcurvelynx/pull/623?

src/lightcurvelynx/graph_state.py
```diff
@@ -502,13 +503,19 @@ def extract_single_sample(self, sample_num):
...
             new_state.states[node_name] = {}
             for var_name, value in self.states[node_name].items():
                 if self.num_samples == 1:
                     new_state.states[node_name][var_name] = value
                 else:
                     new_state.states[node_name][var_name] = value[sample_num]
+
+        # Copy over the fixed vars information.
+        for node_name, var_list in self.fixed_vars.items():
+            new_state.fixed_vars[node_name] = var_list.copy()
```
hombit: Same question about list elements here

src/lightcurvelynx/graph_state.py
jeremykubica: They are actually sets of strings (variable names). In general it shouldn't be modified and a copy would be fine. But to be safe, I will change to deepcopy.

src/lightcurvelynx/graph_state.py
jeremykubica: Yes. It is technically copying meta-data that was added in #625 (merged today). That PR was laying the ground work for a few different approaches, including #623 (but useful outside that too)

PR #645, opened 2025-12-10
Notebook update

jeremykubica:
Update the notebooks based on feedback from Steve Schulze.

### DIFF
--- a/docs/notebooks/adding_models.ipynb
+++ b/docs/notebooks/adding_models.ipynb
@@ -69,8 +69,8 @@
     "    \"\"\"\n",
     "\n",
     "    def __init__(self, brightness, frequency, **kwargs):\n",
-    "        # The call the super's init sets up the standard physical model\n",
-    "        #  parameters, such as ra, dec, redshift, etc.\n",
+    "        # The call to the super's init sets up the standard physical model\n",
+    "        # parameters, such as ra, dec, redshift, etc.\n",
     "        super().__init__(**kwargs)\n",
     "        self.add_parameter(\n",
     "            \"brightness\",\n",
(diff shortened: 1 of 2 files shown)


docs/notebooks/adding_models.ipynb
```diff
@@ -69,6 +69,8 @@
     "    \"\"\"\n",
     "\n",
     "    def __init__(self, brightness, frequency, **kwargs):\n",
+    "        # The call the super's init sets up the standard physical model\n",
+    "        #  parameters, such as ra, dec, redshift, etc.\n",
```
olivialynn:
```suggestion
    "        # The call to the super's init sets up the standard physical model\n",
    "        # parameters, such as ra, dec, redshift, etc.\n",
```

docs/notebooks/adding_models.ipynb
```diff
@@ -110,7 +112,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As described above, the `__init__()` function adds two parameters `brightness` and `frequency`. These are set from the input arguments of the same name, but, as we will see below, could be set by a variety of inputs. In each parameterized node inherits the parameters set by its parent node. For the `BasePhysicalModel` class this includes: `dec`, `distance`, `ra`, `redshift`, and `t0`.\n",
+    "As described above, the `__init__()` function adds two parameters `brightness` and `frequency`. These are set from the input arguments of the same name, but, as we will see below, could be set by a variety of inputs. In each parameterized node inherits the parameters set by its parent node. These are set up in the call `super().__init__(**kwargs)`. For the `BasePhysicalModel` class these parameters include: `dec`, `distance`, `ra`, `redshift`, and `t0`.\n",
```
olivialynn:
```suggestion
    "As described above, the `__init__()` function adds two parameters `brightness` and `frequency`. Here they are set from the input arguments of the same name; however, as we will see below, they could be set by a variety of inputs. Each parameterized node inherits the parameters set by its parent node, and these parameters are set up in the call `super().__init__(**kwargs)`. For the `BasePhysicalModel` class, these parameters include: `dec`, `distance`, `ra`, `redshift`, and `t0`.\n",
```
nit - potential changes for readability

docs/notebooks/adding_models.ipynb
```diff
@@ -338,7 +340,7 @@
     "\n",
     "The heart of each model is its `compute_sed()` function, which determines how the node is actually simulated.\n",
     "\n",
-    "While the examples above are relatively simplistic (a sin wave or a constant output), we can made arbitrarily complicated models. We suggest that developers of new models to look at the examples in `src/lightcurvelynx/sources/` for examples."
+    "While the examples above are relatively simplistic (a sin wave or a constant output), we can made arbitrarily complicated models. We suggest that developers of new models to look at the examples in `src/lightcurvelynx/models/` for examples."
```
olivialynn:
```suggestion
    "While the examples above are relatively simplistic (a sin wave or a constant output), we can made arbitrarily complicated models. We suggest that developers of new models look at the examples in `src/lightcurvelynx/models/` for further demonstration."
```

olivialynn (approved):
Looks good!

Also added some potential wording changes now that I'm looking through existing things with fresher eyes.

PR #652, opened 2025-12-17
Make extrapolation more general

mi-dai:
Make extrapolation more general so it can fit on multiple points instead of using only the last point.

### DIFF
--- a/src/lightcurvelynx/models/physical_model.py
+++ b/src/lightcurvelynx/models/physical_model.py
@@ -439,10 +439,7 @@ class SEDModel(BasePhysicalModel):
                 # the list of wavelengths to extrapolate.
                 valid_mask = query_waves >= min_valid_wave
                 before_wave_queries = query_waves[~valid_mask]
-                if hasattr(self._wave_extrap_before, "nfit"):
-                    n_select_wave_before = self._wave_extrap_before.nfit
-                else:
-                    n_select_wave_before = 1
+                n_select_wave_before = self._wave_extrap_before.nfit
                 query_waves = np.concatenate(
                     (min_valid_wave + 10.0 * np.arange(n_select_wave_before), query_waves[valid_mask])
                 )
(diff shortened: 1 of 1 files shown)


src/lightcurvelynx/utils/extrapolate.py
```diff
@@ -92,34 +96,51 @@ class ZeroPadding(FluxExtrapolationModel):
...
         -------
         flux : numpy.ndarray
             A N x M matrix of extrapolated values. Where M is the number of query points and
             N is the number of flux values at the last valid point.
         """
-        N_len = len(last_flux)
+
+        if extrap_axis not in ["time", "wavelength"]:
+            raise ValueError("Invalid extrap_axis value: only 'time' or 'wavelength' is valid")
+        if not np.isscalar(last_values):
```
jeremykubica: Can you add a comment on what this block is doing? I'm not sure I follow the logic of every path.

src/lightcurvelynx/utils/extrapolate.py
```diff
@@ -279,37 +353,185 @@ def __init__(self, rate):
...
+    """Linear extrapolation based on a linear fit to the last few points.
+
+    Attributes
+    ----------
+    nfit : float
+        The number of points to perform the linear fit on.
+    """
+
+    def __init__(self, nfit=5):
+        self.nfit = nfit
```
jeremykubica: If we do make `nfit` an attribute of the base class, we would need to reverse the order of the two statements here in order to set a default and then override it.

src/lightcurvelynx/utils/extrapolate.py
```diff
@@ -92,34 +96,51 @@ class ZeroPadding(FluxExtrapolationModel):
...
         flux : numpy.ndarray
             A N x M matrix of extrapolated values. Where M is the number of query points and
             N is the number of flux values at the last valid point.
         """
-        N_len = len(last_flux)
+
+        if extrap_axis not in ["time", "wavelength"]:
+            raise ValueError("Invalid extrap_axis value: only 'time' or 'wavelength' is valid")
+        if not np.isscalar(last_values):
+            if np.any(last_values < query_values):
```
jeremykubica: To confirm my understanding, this statement is determining whether it is extrapolating before or after. Is that right?

Would it be easier to just add a "before/after" boolean to the extrapolate call?

jeremykubica (changes requested):

src/lightcurvelynx/utils/extrapolate.py
mi-dai: Also not needed.

PR #684, opened 2026-01-12
Add the option to filter to rest frame time bounds

### LINKED ISSUE #676: Add ability to specify time windows in rest frame
Currently the simulation step takes the time windows in observer frame. Add a second parameter (tuple of start and end) that does the pruning in the rest frame (accounting for redshift).

jeremykubica:
Closes #676

Users can limit samples to only times in a window around `t0` in either the observer frame or rest frame, but not both.


hombit (approved):
Looks good!

PR #720, opened 2026-02-10
Fix pickling of BasicMathNode

jeremykubica:
BasicMathNode failed to pickle because it was storing the module and a lambda as attributes. This deletes them during dump and reinstates them during load, allowing the object to be pickled.

It also simplified the code by replacing the function created in the constructor with the object's `eval` method. This second change does not modify behavior.


hombit (approved):
Looks good! It may be clearer to implement `__getnewargs_ex__` and make `__getstate__` return empty state, and `__setstate__` to do nothing.

PR #728, opened 2026-02-12
Fix NumpyRandomFunc with multi-dimensional parameters

jeremykubica:
The `NumpyRandomFunc` nodes were failing when given non-scalar parameters and multiple samples, such as using a multivariate normal distribution due to broadcasting problems. This PR fixes the problem by improving the broadcasting rules.


hombit (approved):
Looks good, and I couldn't find a better solution for broadcasting

PR #741, opened 2026-02-20
Add a default override for LSST exposure time

jeremykubica:
The parquet version of the CCD visit table does not always include the exposure time. If it is not present, fall back to the default of 30.0 (15.0 + 15.0)


hombit (approved):
1×30s is consistent with DP1, let's keep it!

PR #752, opened 2026-03-04
Add ability to squash pooch output

### LINKED ISSUE #750: Too many Pooch warnings
**Feature request**

I'm getting a lot of warnings from Pooch with file hash values when running `PassbandGroup.from_preset(preset="LSST")`. This information is not useful for users because we don't provide a way of using these hashes. Let's get rid of them or just inform that we've downloaded a file and put it somewhere.

**Before submitting**
Please check the following:

jeremykubica:
Closes #750 

This PR adds multiple options for squashing output in order to silence Pooch's warnings about the SHA hash during download.
- `SquashLogging` is a context manager that temporarily changes the logging output level of a given logger. This is what actually squashes the Pooch messages.
- `SquashOutput` is extended to have the ability to redirect the stdout/stderr to logs. I added this when I thought Pooch was using stdout for the messages. Although this is not needed for the Pooch change, it is useful for general output squashing (without dropping the output altogether).

### DIFF
--- a/src/lightcurvelynx/utils/data_download.py
+++ b/src/lightcurvelynx/utils/data_download.py
@@ -10,7 +10,12 @@ from lightcurvelynx.utils.io_utils import SquashLogging
 logger = logging.getLogger(__name__)
 
 
-def download_data_file_if_needed(data_path, data_url, force_download=False):
+def download_data_file_if_needed(
+    data_path,
+    data_url,
+    force_download=False,
+    silent=False,
+):
     """Download a data file from a URL and save it to a specified path.
 
     Parameters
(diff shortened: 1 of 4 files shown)


hombit (commented):
>> Thank you! Would it possible to keep reports about the downloads (URL -> local path), but skip hashes, etc?

The LightCurveLynx code already logs that information to info. This should just shut off the sha logging (which I confirmed with new tests and manual notebook runs). If you just rerun the notebooks they won't show up for one of two reasons.
1) The files are downloaded by a previous run.
2) LightCurveLynx logs at the WARNING level by default.
I'm not sure the best approach for the second point. We could have those be print statements or change the default logging level. Any preferences?

hombit (approved):

PR #757, opened 2026-03-05
Fix the extrapolation for redback models

jeremykubica:
As an intermediate step, the `RedbackWrapperModel` computes and queries a `RedbackTimeSeriesSource`, which uses a 2d spline to represent the model. This spline is only valid within the range of times and wavelengths on which it was created. Extrapolation outside this range can lead to non-physical results.

This PR precomputes the `RedbackTimeSeriesSource` and uses its bounds for LightCurveLynx's extrapolation. The user can then provide custom extrapolation functions like any other model. If they don't and query points are outside the range, they will get a warning (already handled by the superclass).


hombit (approved):
I think it should be good, but I haven't found a simple reproducer of the issue it fixes...

PR #758, opened 2026-03-05
Fix test warning

jeremykubica:
We were getting test warnings because matplotlib was opening too many plots without closing them. This PR adds logic to close the plots after each test.


olivialynn (approved):
Interesting! Good to know 👀

PR #771, opened 2026-03-12
Issue769

jeremykubica:
Closes #769 

Replace the internal downloader for SVO with astroquery. This allows us to also add an option to download all filters for the combination of facility and instrument. The PR also changes the terminology (e.g. observatory->facility) to be consistent with astroquery.


hombit (approved):
Thank you

PR #796, opened 2026-03-27
Extend default color map

jeremykubica:
Closes #795 

Extend the color map when we have >7 filters.


hombit (approved):
I approve it, but while we are changing this code, we have a good chance to improve it.  
Let's use tab10 for a small amount of colors. It is more consistent with tab20, and I believe it is more colorblind-friendly than "bgrcmyk."

PR #803, opened 2026-04-14
Data dir

### LINKED ISSUE #802: Change `_LIGHTCURVELYNX_BASE_DATA_DIR`
**Feature request**

We should change the default data directory path to move it out of the package installation path. The package installation path is not a safe choice for shared environments (e.g. NERSC's desc-td environment), read-only filesystems, and other edge cases, such as [importing packages from zip archives](https://docs.python.org/3/library/zipimport.html). Also each new installation would cause data re-downloading.

I think that it is reasonable to follow best practices for Unix packages and 1) use `$XDG_CACHE_HOME` (with fallback to `$USER/.cache`), 2) allow changing the cache directory with a config file at `$XDG_CONFIG_HOME/lightcurvelynx.conf` (with fallback to `$USER/.config/lightcurvelynx.conf`).

I'm not sure about Windows, but I think for most people it would be something like `%LOCALAPPDATA%\LightCurveLynx\Cache`. It looks like Astropy is doing something like that.

**Before submitting**
Please check the following:

mi-dai:
## Solution Description

Changed _LIGHTCURVELYNX_BASE_DATA_DIR ot _LIGHTCURVELYNX_DOWNLOAD_DATA_DIR. Current base dir is kept for bayesn and roman files. Allow override using $LIGHTCURVELYNX_DATA_DIR.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,6 +1,33 @@
 Frequently Asked Questions
 ========================================================================================
 
+Where Does LightCurveLynx Store Downloaded Data?
+-------------------------------------------------------------------------------
+
+LightCurveLynx downloads data files (e.g., OpSim tables, passbands, dust maps) on demand
+and caches them so they do not need to be re-downloaded on subsequent runs. The cache
+location is resolved in the following priority order:
+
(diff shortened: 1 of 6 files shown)


jeremykubica (approved):
This is great! Thank you for putting it together

PR #809, opened 2026-04-24
Support Flexible Noise Models

### LINKED ISSUE #548: Make a more flexible noise calculation interface for custom surveys
**Feature request**

There are a lot of different inputs possible for the noise model. We should try to make an interface which:  
1. Would help users understand the inputs for the noise model and what components it has;
2. Would be flexible enough to accept different sets of inputs in different units (e.g. sky background may be in mag per arcsec², per pixel², or in fluxes instead of magnitudes) and derive all necessary inputs for the noise model;
3. Make the model choice very explicit.

**Before submitting**
Please check the following:

jeremykubica:
Closes #548 

This PR changes how noise is computed. Instead of using a function `bandflux_error_point_source` written into each `ObsTable` subclass, we define a noise model that is used by the simulation code. This noise model needs to know how to access each ObsTable's fields, so specializations will be needed for different ObsTables. However it also allows general noise models such as the constant noise model and (in the future) a learned noise model.


src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -0,0 +1,216 @@
+from abc import ABC, abstractmethod
+
+import numpy as np
+from lightcurvelynx.noise_models.noise_utils import poisson_bandflux_std
+
```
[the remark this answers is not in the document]
jeremykubica: Done

src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -0,0 +1,217 @@
...
+
+class FluxNoiseModel(ABC):
+    """An abstract baseclass noise model for simulating bandflux measurements."""
+
+    @abstractmethod
+    def __init__(self):
+        pass
+
```
hombit: I think we don't need it to be abstract and don't even need it here
```suggestion
```

src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -0,0 +1,217 @@
...
+        noisy_bandflux = rng.normal(loc=bandflux, scale=self.noise_level)
+        return noisy_bandflux, np.full_like(bandflux, self.noise_level, dtype=float)
+
+
+class PoissonFluxNoiseModel(FluxNoiseModel):
+    """A noise model that simulates photon noise for bandflux measurements
+    with a Poisson noise level that are extracted from an ObsTable.
+
+    This class is meant to be subclassed for specific ObsTable implementations
+    where the columns vary. Subclass should override the compute_flux_error method.
```
hombit: Should we make `compute_flux_error` abstract then? We still can call it over super() when needed

src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -0,0 +1,217 @@
...
+            obs_table=obs_table,
+            indices=indices,
+        )
+
+        # Make sure the array is a numpy array.
+        flux_err = np.asarray(flux_err)
+
+        # Generate the actual noisy bandflux measurements.
+        if rng is None:
+            rng = np.random.default_rng()
```
hombit:
```suggestion
        rng = np.random.default_rng(rng)
```

src/lightcurvelynx/obstable/fake_obs_table.py
```diff
@@ -2,13 +2,54 @@
...
-from lightcurvelynx.astro_utils.noise_model import poisson_bandflux_std
+from lightcurvelynx.noise_models.base_noise_models import PoissonFluxNoiseModel
+from lightcurvelynx.noise_models.noise_utils import poisson_bandflux_std
 from lightcurvelynx.obstable.obs_table import ObsTable
 from lightcurvelynx.obstable.obs_table_params import ParamDeriver
 
 logger = logging.getLogger(__name__)
 
 
+class FakeObsTableFluxNoiseModel(PoissonFluxNoiseModel):
```
hombit: Maybe `FakeObsTablePoissonNoise`?

hombit (approved):
Thanks! I didn't check the math, assuming that it was not changed here.

src/lightcurvelynx/noise_models/base_noise_models.py
jeremykubica: No. Because `PoissonFluxNoiseModel` is not meant to be an abstract class, but rather the default implementation. Subclasses will have different implementations that should override the function.

src/lightcurvelynx/obstable/fake_obs_table.py
jeremykubica: Done

mi-dai (approved):
Looks good to me. I think we've discussed to separate noise model from obstable in the future?

PR #812, opened 2026-04-28
Support PZFlow noise models

### LINKED ISSUE #805: Introduce data-driven noise simulation
**Feature request**

Many of our users are interested in data-driven noise simulation. The goal is to sample flux error conditionally to perfect flux and obstable metadata. This may be done with the following steps:
1. Select training data (flux, flux error, passband, and other observational metadata excluding visit ID and time), stratified by magnitude/flux. I guess we need order of 1—10 million "sources". This may be done with LSDB using a HATS catalog.
2. Train a model with pzflow or similar.
3. Make an interface to use the trained model.

We can start with Rubin DP1 catalog as a training set. We should publish the result model. We can also do ZTF, but it is soft-blocked by #643

Flux error reliability in the catalog data is out of the scope of this issue, see https://github.com/lincc-frameworks/uncle-val for the related work.

**Before submitting**
Please check the following:

jeremykubica:
Closes #805.

Create a noise model type that loads and samples from a pzflow model. Also includes tests and an example notebook.


docs/notebooks/pre_executed/pzflow_noise_models.ipynb
```diff
@@ -0,0 +1,248 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# pzflow learned noise models\n",
+    "\n",
+    "In this notebook we show how we can use the [pzflow package](https://jfcrenshaw.github.io/pzflow/) to approximate a bandflux noise model. The pzflow model is a model of the standard deviation of the noise given the input bandflux and other relevant parameters from the `ObsTable`.\n",
```
hombit:
```suggestion
    "In this notebook we show how we can use the [pzflow package](https://jfcrenshaw.github.io/pzflow/) to  sample bandflux errors (uncertainties) from a photometric catalog. The pzflow model, as well as the input catalog data, are assumed to provide the standard deviation of the bandflux, while the noise itself is assumed to be Gaussian. The pzflow model samples bandflux errors conditionally to bandflux, passband and other relevant parameters from the `ObsTable`.\n",
```

docs/notebooks/pre_executed/pzflow_noise_models.ipynb
```diff
@@ -0,0 +1,248 @@
...
+    "from lightcurvelynx.noise_models.pzflow_noise_model import PZFlowNoiseModel"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Training the PZFlow model\n",
+    "\n",
+    "We start by generating a fake data set consisting of zero point and readout noise information."
```
hombit:
```suggestion
    "We start by generating a fake observational data set consisting of zero point and readout noise information."
```

docs/notebooks/pre_executed/pzflow_noise_models.ipynb
hombit, line 1: I tried to make it clear that input data is just measurements made by a time domain survey, but maybe we can make it even cleaner with both better description and data frame example "simulating" a few non-variable light curves.

hombit (commented):
One of the big questions is if we need to normalize bandflux_err values, because we may have a huge dynamic range, and neural networks usually hate that. Maybe do something like min-max normalization of log10(bandflux_err)? And the same about flux.

jeremykubica:
> One of the big questions is if we need to normalize bandflux_err values, because we may have a huge dynamic range, and neural networks usually hate that. Maybe do something like min-max normalization of log10(bandflux_err)? And the same about flux.

What do you think would be the best approach here? 1) always use a `log10` normalization, 2) have a default normalization and a Boolean of whether that normalization was used, 3) allow the user to pass a normalization / denormalization function, or 4) something else?

hombit (approved):

PR #819, opened 2026-05-06
Use column transforms for SkyMapper noise model

jeremykubica:
Instead of defining a custom `SkyMapperPoissonFluxNoiseModel`, add a few columns during load so that we can use a standard `PoissonFluxNoiseModel`.


hombit (approved):
Looks good

PR #820, opened 2026-05-06
Change the interface for passing noise models and passbands

jeremykubica:
This is a change to the simulation API. It creates a new `SurveyInfo` object that wraps the `ObsTable`, `PassbandGroup`, and noise model. In addition:
- The noise model is no longer stored in the ObsTable.
- `SurveyInfo` can add default noise models or passband groups given a survey name.
- `SurveyInfo` checks that the noise model, table, and passbands are all mutually compatible.

That majority of this PR is updating the example notebooks.


docs/notebooks/pre_executed/eztaox_example.ipynb
mi-dai: These need to be updated?

src/lightcurvelynx/obstable/obs_table.py
```diff
@@ -223,17 +218,25 @@ def copy(self):
...
+    @property
+    def default_noise_model(self):
+        """Return the default noise model for this ObsTable, if it exists."""
+        return None  # No default noise model.
+
+    @property
+    def default_passband_group(self):
+        """Return the default passband group for this ObsTable, if they exist."""
+        return None  # No default passband group.
+
```
mi-dai: I feel adding default noise model and default passband group in obstable is a bit confusing to users. If we switch to survey info, we should only use survey info instead of using obstable as a short cut.

src/lightcurvelynx/simulate.py
```diff
@@ -587,10 +582,13 @@ def simulate_lightcurves(
         to the result columns.
     num_samples : int
         The number of samples.
-    obstable : ObsTable or List of ObsTable
-        The ObsTable(s) from which to extract information for the samples.
-    passbands : PassbandGroup, Spectrograph, or List of (PassbandGroup, Spectrograph)
-        The passbands to use for generating the bandfluxes or spectra.
+    survey_info : SurveyInfo, ObsTable, List of SurveyInfo, or List of ObsTable
```
mi-dai, line 588: Same comment as above about allowing obstable as survey info

mi-dai (commented):
I think default passband and default noise should be at survey info level instead of in the obstable. How about something like `SurveyInfo.from_default(obstable, surveyname="LSST")` and lookup default passband and noise model given surveyname inside SurveyInfo?)

jeremykubica:
As discussed via slack, I moved the default logic out of the `ObsTable` classes. I left the ability to pass a list of `ObsTable` with a deprecation warning. Can you take another look?

docs/notebooks/pre_executed/eztaox_example.ipynb
```diff
@@ -62,16 +62,16 @@
    "source": [
     "## Create the Survey Data\n",
     "\n",
-    "There are two types of data files we need to run a LightCurveLynx simulation an `ObsTable` and passband information. In most cases we will want to load the simulated or actual observation tables from various surveys.\n",
+    "First we will load the survey information, which consists of the observations (``obstable``), passband information, and noise model. In this example, we create a fake `OpSim` that observes the exact same location on the sky (45.0, -20.0) three times a night, once in each of the 'r\", \"g\", and 'i' bands. This is particularly realistic for a survey, but provides densely sampled output curves for the purpose of this notebook. For examples of more realistic RA/dec distributions and opsims, see the other notebooks.\n",
```
mi-dai: particularly **un**realistic?

docs/notebooks/pre_executed/eztaox_example.ipynb
```diff
@@ -62,16 +62,16 @@
...
     "\n",
-    "In this example we will use the default Rubin passbands. But we will create a fake observation cadence that is better for visualizing microlensing events. Specifically, we will take make two fake `OpSim` that observes the exact same location on the sky (45.0, -20.0) three times a night, once in each of the 'r\", \"g\", and 'i' bands. This is particularly realistic for a survey, but provides densely sampled output curves for the purpose of this notebook. For examples of more realistic RA/dec distributions and opsims, see the other notebooks.\n",
+    "Note that we specify a lot of noise parameters with the fake `OpSim`, but in normal cases users will be loading a predefined opsim with all of that information pre-populated.\n",
     "\n",
-    "Note that we specify a lot of noise parameters with the fake `OpSim`, but in normal cases users will be loading a predefined opsim with all of that information pre-populated."
+    "Since we are using a standard survey type (LSST), we can just pass the name to the `SurveyInfo` info constructor and have it load the correct passbands and noise model."
```
mi-dai: correct -> default

mi-dai (commented):
Looks good! 

I think I saw this sentence in a few places but lost track. "Since we are using a standard survey type (LSST), we can just pass the name to the `SurveyInfo` info constructor and have it load the correct passbands and noise model." I suggest change correct to default.

jeremykubica:
I made the suggested changes. Can you take another look?

mi-dai (approved):

PR #822, opened 2026-05-07
Fix broken test

jeremykubica:
One of the pzflow tests is failing due to a change in the API. Replace the constructor with `file` argument with the `from_file` class method.


hombit (commented):
Do we need to constrain PzFlow version? I would also encourage a patch version release with either this change, or with pinning PzFlow to a previous version

jeremykubica:
> Do we need to constrain PzFlow version? I would also encourage a patch version release with either this change, or with pinning PzFlow to a previous version

Done

hombit (approved):

PR #823, opened 2026-05-07
Prefactor the noise functions

jeremykubica:
A prefactoring to help with future changes to the noise function.

Rename `_assign_zero_points` to `_derive_noise_columns` and include other derivations there as needed. Also adds an optional list of required columns where the code will fail if they are missing.

Add information about the required columns to the noise models including both a list of required noise columns (could be fed into `_derive_noise_columns`) and a compatibility checking function.

Update the sky mapper obstable to account for the two changes above.


src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -48,6 +56,32 @@ def apply_noise(
...
+        -------
+        bool
+            True if the noise model is compatible with the ObsTable, False otherwise.
+        """
+        missing_columns = [col for col in self._required_values if col not in obs_table]
+        if missing_columns:
+            if fail_on_incompatible:
+                raise ValueError(
+                    f"Noise model {self.__class__.__name__} is not compatible with the given ObsTable. "
+                    f"Missing required columns: {missing_columns}"
```
hombit, line 80: Maybe add some actionable advice to the end of the error message?

hombit (approved):

src/lightcurvelynx/noise_models/base_noise_models.py
jeremykubica: I'm going to keep it as-is for now until I have a better idea of what good actionable advice would be. But adding it is a good call.

PR #826, opened 2026-05-07
Clean up the FakeObsTable

jeremykubica:
Redo the `FakeObsTable` so that instead of trying to exhaustively derive columns that could be useful, it uses a given noise model. Users are now responsible for providing the columns that match the noise model given. This is safe because we are not trying to infer what each column means. In addition it should be simpler to use and use less memory.

As part of this, the PR ports over two types of noise models that were previously supported by the `FakeObsTable`:
- `GivenNoiseModel` will use a given noise whether it is per-row, a mapping of filter to noise level, or a single constant.
- `FiveSigmaDepthNoiseModel` is the model we ported for the redback team during the incubator.


src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -247,3 +247,138 @@ def apply_noise(
...
+        noisy_bandflux = rng.normal(loc=bandflux, scale=flux_err)
+        return noisy_bandflux, flux_err
+
+
+class FiveSigmaDepthNoiseModel(FluxNoiseModel):
+    """A noise model that simulates photon noise from only the five-sigma depth information.
+
+    It requires two pieces of information from the ObsTable:
+    - five_sigma_depth: The five-sigma depth in magnitudes.
+    - bandflux_ref: A reference bandflux in the same units as the input bandflux
```
mi-dai, line 319: Talked with @hombit and agreed that `bandflux_ref` will cause confusion, because it indicates we can support different magnitude systems, which we currently don't. We propose to merge this PR and revisit in another issue.

mi-dai (approved):

PR #830, opened 2026-05-08
Document units used by noise models

jeremykubica:


src/lightcurvelynx/noise_models/base_noise_models.py
```diff
@@ -140,8 +140,15 @@ class PoissonFluxNoiseModel(FluxNoiseModel):
...
-    where the columns vary. Subclasses should override the compute_flux_error method.
+    This noise model uses the following values from the ObsTable:
+    - dark_current: Mean dark current (electrons per pixel per second).
+    - exptime: The total exposure time for the observation (seconds).
+    - nexposure: The number of exposures (optional, default is 1).
+    - psf_footprint: Point spread function effective area (pixel^2).
+    - read_noise: Standard deviation of the readout electrons per pixel per exposure.
+    - sky_bg_e: Sky background (electrons / pixel^2).
+    - zp: The photometric zero point (nJy / electron).
+    - zp_err_mag: The uncertainty in the photometric zero point in magnitudes (optional, default is 0.0).
```
mi-dai, line 163: What do users do if their values are in different units? Where do they do the conversion?

jeremykubica: Good question. I added a note about that to this docstring and put a new docstring at the topic with the overall logic for these noise models.

mi-dai (approved):

PR #833, opened 2026-05-12
Fix broken indexing after filtering with a mask

jeremykubica:
The ObsTable does not correctly update the indices after mask filtering, which can cause `get_value_per_row` to fail.

### DIFF
--- a/src/lightcurvelynx/obstable/argus_obstable.py
+++ b/src/lightcurvelynx/obstable/argus_obstable.py
@@ -106,7 +106,7 @@ class ArgusHealpixObsTable(ObsTable):
         # The table uses the healpix IDs as the index. We bring this into its own column for easier access
         # and reset the indices.
         if "healpix" in table.columns:
-            table.reset_index(inplace=True)  # Just reset the index if healpix is already a column.
+            table = table.reset_index()  # Just reset the index if healpix is already a column.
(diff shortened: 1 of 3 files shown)


src/lightcurvelynx/obstable/argus_obstable.py
```diff
@@ -106,9 +106,9 @@ def __init__(
         # The table uses the healpix IDs as the index. We bring this into its own column for easier access
         # and reset the indices.
         if "healpix" in table.columns:
-            table = table.reset_index()  # Just reset the index if healpix is already a column.
+            table.reset_index(inplace=True)  # Just reset the index if healpix is already a column.
```
hombit: This change is very suspicious, why does it make the difference?

hombit (approved):
From my understanding the modern pandas discourages the use on `inplace=True`, and I think Doug checked that it is not improving the performance.

src/lightcurvelynx/obstable/argus_obstable.py
jeremykubica: It doesn't I just thought it looked cleaner. I reverted it.

jeremykubica:
I added another quick update, can you have another look?

hombit (approved):

PR #840, opened 2026-05-13
Add more ObsTable tests

jeremykubica:
Cover more edge cases for the ObsTable (especially around filtering and retrieval). Also adds subsampled version of the DP1 CCD visit table that can be used for both tests and demo notebooks.


tests/lightcurvelynx/data/fake_dp1_ccdvisit.parquet
hombit, line 1: DP1 visits are open for the world, so we can subsample the real one instead of faking.

hombit (approved):

tests/lightcurvelynx/data/fake_dp1_ccdvisit.parquet
jeremykubica: Done

PR #842, opened 2026-05-13
Update pyproject.toml

jeremykubica:
Now that the JAX versions have been updated we can include eztaox in the dev dependencies, which enables the tests in the CI.


jeremykubica:
The JAX versions are still incompatible

PR #845, opened 2026-05-15
Add a test for reading the LSST SV table

jeremykubica:
Add a subsampled version of the SV visit table to the test data and a test that we can correctly read this data from a .db file.


docs/notebooks/pre_executed/rubin_ccdvisit.ipynb
```diff
@@ -19,7 +19,7 @@
    "source": [
     "## Get the Visits Table\n",
     "\n",
-    "Users will need to download the CCDVisits table. Here we use the parquet version of the [DP1 ccdvisits table on LSDB.io](https://data.lsdb.io/hats/dp1/). This notebook assumes the table has been downloaded to a local directory."
+    "Users will need to download the CCDVisits table. Here we use the parquet version of the [DP1 ccdvisits table on LSDB.io](https://data.lsdb.io/hats/dp1/). This notebook assumes the table has been downloaded to a local directory. A subsampled copy of this data is also included locally in \"tests/lightcurvelynx/data/dp1_ccdvisit_subsampled.parquet\". "
```
hombit: Is it actually being distributed with the wheel? I don't think so...

```suggestion
    "Users will need to download the CCDVisits table. Here we use the parquet version of the [DP1 ccdvisits table on LSDB.io](https://data.lsdb.io/hats/dp1/). This notebook assumes the table has been downloaded to a local directory. A subsampled copy of this data is also included in the lightcurvelynx repository at: \"tests/lightcurvelynx/data/dp1_ccdvisit_subsampled.parquet\". "
```

hombit (approved):

PR #846, opened 2026-05-15
Redback Updates

jeremykubica:
A few small updates to the redback models:
- Fix the broken notebook, but only simulating for `t0, t0+50`
- Add a more meaningful error message when redback fails (usually due to bad parameters).
- Add a note on how to automatically load and sample from redback priors.


hombit (approved):
Very nice, thanks

PR #852, opened 2026-05-21
Sort each light curve by time

jeremykubica:
Sort the light curve data in the nested column by time.

### DIFF
--- a/src/lightcurvelynx/simulate.py
+++ b/src/lightcurvelynx/simulate.py
@@ -393,6 +393,10 @@ def _simulate_lightcurves_batch(simulation_info):
     ):
         total_num_obs = 0
 
+        # Create an object-specific nested dictionary for this so we can sort by time
+        # (over all surveys) before appending to the overall nested dictionary.
+        object_nested_dict = {key: [] for key in nested_dict}
+
         for survey_idx in range(num_surveys):
             # Find the indices and times where the current model is seen.
             obs_index = np.asarray(all_obs_matches[survey_idx][idx])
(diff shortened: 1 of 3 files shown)


src/lightcurvelynx/simulate.py
```diff
@@ -458,22 +462,22 @@ def _simulate_lightcurves_batch(simulation_info):
...
                 # any needed ObsTable columns.
-                nested_dict["mjd"].extend(list(obs_times))
-                nested_dict["filter"].extend(list(obs_filters))
-                nested_dict["flux_perfect"].extend(list(bandfluxes_perfect))
-                nested_dict["flux"].extend(list(bandfluxes))
-                nested_dict["fluxerr"].extend(list(bandfluxes_error))
-                nested_dict["survey_idx"].extend([survey_idx] * nobs)
-                nested_dict["is_saturated"].extend(list(saturation_flags))
-                nested_dict["obs_idx"].extend(list(obs_index))
+                object_nested_dict["mjd"].extend(list(obs_times))
```
hombit: Going back and forth between arrays and lists is not super-efficient, we can collect a list or arrays and `np.concatenate` them later.

src/lightcurvelynx/simulate.py
```diff
@@ -485,11 +489,18 @@ def _simulate_lightcurves_batch(simulation_info):
...
 
             # Regardless of the integrator (instrument), we count the number of observations.
             total_num_obs += nobs
 
+        # Sort the object nested dictionary by time (across all surveys) before appending to the
+        # overall nested dictionary. We only do this if we have at least one non-spectral observation.
+        if len(object_nested_dict["mjd"]) > 0:
+            sorted_indices = np.argsort(object_nested_dict["mjd"], kind="stable")
+            for key in nested_dict:
+                nested_dict[key].extend(np.array(object_nested_dict[key])[sorted_indices].tolist())
```
hombit: it looks like we always extend an empty array here with a single new array? Cannot we just assign it?

hombit (approved):

src/lightcurvelynx/simulate.py
jeremykubica: Good call. I made that change.

src/lightcurvelynx/simulate.py
jeremykubica: I added a comment to clarify why not.

hombit (approved):

PR #853, opened 2026-05-21
Small Optimization for GraphState

jeremykubica:
Currently we are creating a bunch of empty sets in the GraphState for unfixed variables. This only creates a set if the node has at least one fixed variable. It should save a little memory and computation time.

### DIFF
--- a/src/lightcurvelynx/graph_state.py
+++ b/src/lightcurvelynx/graph_state.py
@@ -49,7 +49,8 @@ class GraphState:
         The total number of parameters stored in a single sample within GraphState.
     fixed_vars : dict
         A dictionary mapping the node name to a set of the variable names that
-        are fixed (not changed by resampling) in this GraphState instance.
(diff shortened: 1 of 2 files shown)


olivialynn (approved):
Looks good!

As a not-everyday-user, it took me a minute to go over the significance of a fixed variable as distinct from a simple constant. If these are commonly used, it could be nice to have an entry in the glossary. 

Also, is it an intentional choice for an attempt to change the value of a fixed variable to be silently ignored, as in `test_graph_state_fixed`? I could imagine wanting to ignore clutter if we'd expect this to get triggered often, but just wanted to check.

jeremykubica:
Good questions and some more documentation is needed. The fixed variable is meant for use with automatic differential code only and shouldn't be seen by most users. Users will just use a normal variable with a constant. However some optimization packages might need to optimize over parameters and we need to fix parameters that they cannot change during optimization (such as ra, dec).

PR #859, opened 2026-05-27
Fix Issue 857

### LINKED ISSUE #857: PassbandGroup.from_preset has wrong type annotation
**Bug report**

It says `-> None`, while it should be either `-> Self` or `-> "PassbandGroup"`

jeremykubica:
Closes #857

Fix the incorrect type hint (and add a bunch of missing return type hints).


hombit (approved):
Thanks

PR #866, opened 2026-06-01
Allow programmatic access in 

### LINKED ISSUE #865: Avoid stochasticity when using MultiLightcurveTemplateModel
Hello. From what I can understand, MultiLightcurveTemplateModel samples randomly from the pool of lightcurves given (even when not using .from_lclib_file). Is there a way to avoid this, other than looping through all light curves and simulating them individually?
Incidentally, I suppose giving a t0=None will use each lightcurve's own t0. But is there a way to do the same for the position? If I supply a list of RA and DEC?

Thanks a lot

jeremykubica:
Closes #865 

## Solution Description
Allow programmatic access to the light curves in `MultiLightcurveTemplateModel` through an optional `indices` parameter. Adds a demo notebook showing how this can be used to re-simulate known objects (with given ra/dec).

Also allows a non-stateful `GivenValueList` which is useful for testing and distributed computation.

### DIFF
--- a/src/lightcurvelynx/models/lightcurve_template_model.py
+++ b/src/lightcurvelynx/models/lightcurve_template_model.py
@@ -840,7 +840,7 @@ class MultiLightcurveTemplateModel(BaseLightcurveBandTemplateModel):
                 # Assume it is already a parameter or sampler.
                 indices_sampler = indices
         else:
-            all_inds = [i for i in range(len(lightcurves))]
+            all_inds = np.arange(len(lightcurves))
             indices_sampler = GivenValueSampler(all_inds, weights=weights)
 
         self.add_parameter(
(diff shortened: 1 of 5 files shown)


src/lightcurvelynx/models/lightcurve_template_model.py
```diff
@@ -811,11 +827,25 @@ def __init__(
...
+            if isinstance(indices, list | np.ndarray):
+                indices = np.asarray(indices)
+                if np.any((indices < 0) | (indices >= len(lightcurves))):
+                    raise ValueError("Indices must be between 0 and the number of light curves.")
+                indices_sampler = GivenValueList(indices)
+            else:
+                # Assume it is already a parameter or sampler.
+                indices_sampler = indices
+        else:
+            all_inds = [i for i in range(len(lightcurves))]
```
hombit:
```suggestion
            all_inds = np.arange(len(lightcurves))
```

hombit (approved):

PR #871, opened 2026-06-04
Add ability to manually specify redback bounds

jeremykubica:
The math in the underlying redback models can fail if evaluated out of bounds. In some cases we can infer the bounds from the redback result object (sncosmo object), but other cases the computation will fail before that. This PR allows the user to specify manual time and wavelength bounds in these cases.


hombit (approved):
Thanks

Issue #844, opened 2026-05-14
Gravitationally lensed transient simulation functionality


erinhay:
**Feature request**
Add support for simulating gravitationally lensed transients using functionality from [lenstronomy](https://github.com/lenstronomy/lenstronomy). Inspiration for this feature may come from the [lensed_supernova_simulator_tool](https://github.com/Nikki1510/lensed_supernova_simulator_tool)

To enable the simulation of lensed supernovae or quasars, we need to:
1. Define a LensGalaxy model (which contains info about the mass profile).
2. Calculate image positions, magnifications, and time delays.
3. Apply the time-delay/magnification transformations, with option for "unresolved" (a single light curve consisting of the summed fluxes of all images) vs "resolved" (separate light curves for each image), to a source model using a StrongLens effect.

I'm not exactly sure at what level the lenstronomy objects should be wrapped. Perhaps passing redshift, model_list, model_kwargs to the LensGalaxy model is best and then the LensGalaxy model have a get_lensing_properties(self, source_pos_x, source_pos_y) function that solves the lensing equation for some source position to give the positions, magnifications, and time-delays for each of the resulting images. This way, the user still has control over the lens mass model specification (e.g., SIE, EPL, adding shear effects). There will need to be some way of converting ra, dec to an (x,y) coordinate for the source and lens (I think...). I'm a bit worried about the speed of solving the lens equation and the chance that a source position is not well enough aligned with the lens position to produce multiple images in lens equation solving...

The effect should be fairly straightforward to implement:
```
class StrongLens(EffectModel):
    """An effect that applies strong gravitational lensing effect to the flux density.
    """

    def __init__(self, macro_mag, time_delay, rest_frame=True, **kwargs):
        super().__init__(**kwargs)

        # Define the effect's parameters. Note that these do not need to match
        # the argument names. They should be chosen so that they do not overlap
        # parameters within the model node (e.g. we can't use t0).
        self.add_effect_parameter("macro_mag", macro_mag)
        self.add_effect_parameter("time_delay", time_delay)

        # Set the rest_frame parameter.
        self.rest_frame = rest_frame

    def apply(
        self,
        flux_density,
        times=None,
        wavelengths=None,
        macro_mag=None,
        time_delay=None,
        **kwargs,
    ):
        # Basic error checking to make sure we got values for the parameters we need.
        if macro_mag is None or time_delay is None:
            raise ValueError("macro_mag and time_delay must be provided")
        if times is None:
[...]

PR #875, opened 2026-06-14
Add the ability to expand GraphState

jeremykubica:
This is step 1 of at least three in supporting #844. This PR introduces code that will allow the sampler to receive a list of dictionaries for correlated parameters (different observations of the same lensed source), flatten them, and insert them into the `GraphState`. It expands the rows of the `GraphState` such that each previous row is repeated once for each of the subparameters.

This will be used for the lensing effect to create parameters for dRA, dDEC, and dT0. In later PRs, we will introduce code to apply those offsets during and effect (this will require some new functionality in `BasePhysicalModel`).

### DIFF
--- a/src/lightcurvelynx/graph_state.py
+++ b/src/lightcurvelynx/graph_state.py
@@ -558,7 +558,7 @@ class GraphState:
             repeats = np.asarray(repeats)
             if repeats.ndim != 1:
                 raise TypeError("repeats must be an int or a 1D array-like of integers.")
-            if repeats.dtype.kind not in {"i", "u"}:
+            if not np.issubdtype(repeats.dtype, np.integer):
                 raise TypeError("Entries in repeats must be integers.")
             if len(repeats) != self.num_samples:
                 raise ValueError(
--- a/src/lightcurvelynx/graph_state.py
+++ b/src/lightcurvelynx/graph_state.py
@@ -535,6 +535,64 @@ class GraphState:
             for var_name, value in node_vars.items():
                 self.set(node_name, var_name, value, force_copy=force_copy, fixed=all_fixed)
 
+    def repeat(self, repeats):
+        """Expand the GraphState to a new number of samples by repeating the existing values
+        for all node.parameter combinations.
+
+        Note
+        ----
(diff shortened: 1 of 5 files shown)


src/lightcurvelynx/math_nodes/state_expansion_node.py
```diff
@@ -0,0 +1,166 @@
...
+            raise ValueError("You must provide either 'repeats' or 'param_values' to StateExpansionNode.")
+        else:
+            # No new parameters to add, just repeat the rows.
+            self.new_param_names = []
+
+        super().__init__(
+            func=self._non_func,  # We will override compute() so the function doesn't matter.
+            repeats=repeats,
+            param_values=param_values,
+            outputs=["org_inds", "sub_inds"] + self.new_param_names,
```
hombit, line 80: What if we would repeat twice? Or it could never happen?

src/lightcurvelynx/graph_state.py
```diff
@@ -535,6 +535,67 @@ def update(self, inputs, force_copy=False, all_fixed=False):
...
+            The sum of the entries in this array will determine the new number of samples.
+        """
+        # If we get a scalar integer, repeat all samples by that amount.
+        if isinstance(repeats, int | np.integer):
+            repeats = np.full(self.num_samples, int(repeats), dtype=int)
+        else:
+            repeats = np.asarray(repeats)
+            if repeats.ndim != 1:
+                raise TypeError("repeats must be an int or a 1D array-like of integers.")
+            if repeats.dtype.kind not in {"i", "u"}:
```
hombit:
```suggestion
            if not np.issubdtype(repeats.dtype, np.integer):
```

hombit (approved):

src/lightcurvelynx/math_nodes/state_expansion_node.py
jeremykubica: The way the sampling works, you can only sample from each node once during an iteration. This is done to ensure the statistical consistency of all the items in a single row but also enforces the correct behavior here. You can't call repeat from the same node twice in the same sampling run.

PR #879, opened 2026-06-17
Support sampling from the modeldag package

jeremykubica:
Support using samplers from https://github.com/MickaelRigault/modeldag


hombit (approved):
Looks good!

PR #880, opened 2026-06-17
Handle pre-explosion times in RedbackWrapperModel

nikhil-sarin:
## Summary

Explosion-based redback models (kilonovae, supernovae, etc.) are only defined for `time >= 0` (days since explosion). Previously, if `obs_time_window_offset` included times before the explosion (e.g. `(-5, 20)`), `RedbackWrapperModel` would pass negative shifted times to the source function, causing crashes or garbage flux values.

This PR fixes `compute_sed` to split observations into pre- and post-explosion, evaluate the model only for post-explosion times, and return zero flux for pre-explosion observations. `_do_bounds_precomputation` is also updated to clip to `time >= 0` when determining wavelength bounds.

With this change, users can safely use negative `obs_time_window_offset` starts to simulate realistic pre-explosion non-detections, which will appear as sub-threshold observations in `results_augment_lightcurves`.

## Changes

- `compute_sed`: split `shifted_times` into pre/post-explosion; evaluate source only on post-explosion times; reconstruct full output with zeros for pre-explosion entries
- `_do_bounds_precomputation`: clip to `min_phase = max(0, _min_phase)` before calling source so bounds inference never hits negative times

### DIFF
--- a/src/lightcurvelynx/models/redback_models.py
+++ b/src/lightcurvelynx/models/redback_models.py
@@ -390,12 +390,6 @@ class RedbackWrapperModel(SEDModel, CiteClass):
         # simulated survey, which is the physically correct behaviour.
         post_explosion = shifted_times >= 0.0
         post_times = shifted_times[post_explosion]
-
-        # Determine the output shape: (n_wavelengths, n_times) or (n_times,).
-        if wavelengths.ndim == 0 or np.ndim(wavelengths) == 0:
-            n_wave = 1
-        else:
-            n_wave = len(wavelengths)
         n_time = len(shifted_times)
 
         if post_times.size > 0:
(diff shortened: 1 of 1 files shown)


jeremykubica:
Is this restriction true of all models? Or do you want to allow some of them to have `t0` < 0? In particular SALT-style models use a `t0` at the peak of the explosion, right? 

We ran into this problem ourselves and implemented a different fix in #871, where you can set the phase bounds. We made these settable because of the question above (wanting some models to support negative t0). 

But if that is not the case and you want all models to block `t0 < 0` then you could:
1) either
  a) set the default minimal phase bound to 0 using `phase_bounds=(0.1, None)` on line 82
  b) force `self._min_phase=0` on line 157.
2) set the default time extrapolation to zero padding by adding a line after that on line 83 `time_extrapolation=(ConstantPadding(0.0), None)`

LightCurveLynx automatically truncates the evaluation to only the valid phase bounds prior to compute_sed() function then uses extrapolation to fill in the values. If`self._min_phase` is 0, it should not call compute_sed() with phases less than zero (although in theory users could manually do this if they call compute_sed). 

You can see an example of how we manually pass in the phasebounds here: https://lightcurvelynx.readthedocs.io/en/latest/notebooks/pre_executed/redback_example.html

Of course, setting the defaults only makes sense if we want to block all models from evaluating t0 < 0. Otherwise we should keep it user defined (or ideally tag redback functions with their acceptable range so we can read it automatically).

jeremykubica:
Replaced by #881

PR #881, opened 2026-06-18
Make redback phase bounds required

jeremykubica:
Make redback phase bounds required so the user doesn't accidentally run with invalid bounds.


src/lightcurvelynx/models/redback_models.py
```diff
@@ -49,16 +52,19 @@ class RedbackWrapperModel(SEDModel, CiteClass):
         The redback model's Bilby priors.
     parameters : dict, optional
         A dictionary of parameter setters to pass to the source function.
+    phase_bounds : tuple of (float, float), required
+        A tuple of (min_phase, max_phase) in days to set the phase bounds for the model.
+        This is required because redback models are only defined for certain phase ranges, and we want to
+        make sure we are not evaluating the model at invalid points.
+        Defaults to None which raises an error.
```
mi-dai, line 61: What if users set phase_bounds=(None, None)? Do users have to know the phase bounds for each redback model? Is it documented?

jeremykubica: (None, None) is fine as it means (-inf, inf). We do this for some of the tests.

Yes. The users need to know the bounds for the redback models. These are not documented or programmatically available to my knowledge.

jeremykubica: As discussed over slack, I added more documentation.

mi-dai (approved):

Issue #885, opened 2026-06-23
JOSS REVIEW: minor comments


temuller:
These are comments related to the [JOSS review](https://github.com/openjournals/joss-reviews/issues/10629) of lightcurvelynx.

The documentation is very extensive, detailed and easy to follow. However, make sure that the examples shown don’t have errors as output. See below a non-extensive list of minor things to fix/change.

**Getting Started**
https://lightcurvelynx.readthedocs.io/en/latest/notebooks/introduction.html#PassbandGroup 
Avoid warnings when trying to download already existing passbands

https://lightcurvelynx.readthedocs.io/en/latest/notebooks/introduction.html#Generate-the-simulations
Some errors in the outputs. For instance, `print(lightcurves.iloc[0].lightcurve)` returns `None`, which means that the lightcurves object has missing components causing the errors.

https://lightcurvelynx.readthedocs.io/en/latest/notebooks/introduction.html#Saving-Data 
This example runs locally but needs to be fixed in the docs

https://lightcurvelynx.readthedocs.io/en/latest/notebooks/simple_snia.html 
This example runs locally but needs to be fixed in the docs

**Example Simulations**
https://lightcurvelynx.readthedocs.io/en/latest/notebooks/pre_executed/skymapper_example.html#Filter-Information 
Plotting the filters works but only after some manual fixes, such as manually downloading the filters and setting the file names correctly

**Other Surveys**
https://lightcurvelynx.readthedocs.io/en/latest/notebooks/pre_executed/skymapper_example.html#Generate-the-simulations 
The SkyMapper table has NaN values in the `sky_bg_e` column that causes an error. This works after manually replacing the NaN values, but the author should implement a fix to bypass this type of error (e.g. removing rows with NaN values or replacing them with other values).

jeremykubica:
These were addressed in #886 #887 and #888

PR #886, opened 2026-06-23
Address some tutorial notebook issues

jeremykubica:
Address the first few problems from #885 
- Change the logging level in the intro notebook to WARNING to remove noisy messages
- Uses RA and dec that match data set (which we changed recently) so we do not get Nones in the data
- Use temporary directories for output so this runs on the readthedocs servers
- Add sncosmo to the default installs so it can be used by all notebooks


hombit (approved):
It is good, but we should not forget to update conda package with sncosmo dep!

Issue #891, opened 2026-07-01
JOSS review: Minor comments


charlotteaward:
Here are my minor comments regarding the [JOSS review](https://github.com/openjournals/joss-reviews/issues/10629) of lightcurvelynx. 

In the tutorial notebooks, cell 1 of citations.ipynb gives me an error: `ValueError: 'redshift' parameter is required for the AGN model.`

I think the `pyarrow` dependency might need the addition of version requirement of >19.0.1. When I had an earlier version, I would get the error `OSError: Repetition level histogram size mismatch` when reading in parquet files in the example notebook ccd_obstables.ipynb.

I had a quick question arising from the host_source_models.ipynb example notebook, regarding how overlapping sources are treated. I see that the PSF footprint size can be provided to the noise models and also that PSF FWHM can be provided in the observation tables. Would it be possible to get a quick description in this notebook or elsewhere in the documentation about how PSF models are used, and how the PSF is considered when the code generates photometry of blended sources?

PR #896, opened 2026-07-01
Issue891

jeremykubica:
Address 2.5 of the 3 points in issue #891:
- Fixes the error in the citations notebook by including the missing parameters.
- Pins pyarrow>19.0.1
- Clarifies how `AdditiveMultiObjectModel` adds the flux densities.

We still need a description of how PSF is actually used in the noise models somewhere.

### DIFF
--- a/docs/notebooks/citations.ipynb
+++ b/docs/notebooks/citations.ipynb
@@ -28,6 +28,8 @@
     "    t0=0.0,\n",
     "    blackhole_mass=1e9 * M_SUN_G,\n",
     "    edd_ratio=0.9,\n",
+    "    redshift=0.01,\n",
+    "    distance=100.0,\n",
     "    node_label=\"AGN\",\n",
     ")\n",
     "\n",
@@ -108,7 +110,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "lightcurvelynx",
+   "display_name": "lightcurvelynx (3.13.8)",
    "language": "python",
    "name": "python3"
   },
@@ -122,7 +124,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.4"
+   "version": "3.13.8"
   }
  },
  "nbformat": 4,
--- a/docs/notebooks/citations.ipynb
+++ b/docs/notebooks/citations.ipynb
@@ -20,7 +20,7 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
-    "\n",
+    "from astropy.cosmology import FlatLambdaCDM\n",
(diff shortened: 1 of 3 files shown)


docs/notebooks/citations.ipynb
```diff
@@ -28,6 +28,8 @@
     "    t0=0.0,\n",
     "    blackhole_mass=1e9 * M_SUN_G,\n",
     "    edd_ratio=0.9,\n",
+    "    redshift=0.01,\n",
+    "    distance=100.0,\n",
```
mi-dai: Why both redshift and distance are needed?

docs/notebooks/host_source_models.ipynb
```diff
@@ -70,6 +70,13 @@
     "plt.show()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that `AdditiveMultiObjectModel` simply adds up the rest0frame flux density produced by each model. If you want to account for the specifics of extended sources (variation of flux density with position), that should be computed by the individual models."
+   ]
```
mi-dai, line 78: Is this trying to answer the question about PSF?

docs/notebooks/citations.ipynb
jeremykubica: It either needs redshift+cosmology or redshift and distance. I went with distance so I didn't need to import the cosmology package, but let me just use cosmology to make it clearer.

docs/notebooks/host_source_models.ipynb
jeremykubica: Partially. We don't use PSF when adding sources, so there isn't much to actually say here.

docs/notebooks/citations.ipynb
mi-dai: OK. I was worried that these two numbers doesn't match

docs/notebooks/host_source_models.ipynb
mi-dai: I think the last sentence is confusing. Can we remove it? Also typo in `rest0frame`

mi-dai (commented):
Resubmitting the review to making sure my last comment is seen. "I think the last sentence is confusing. Can we remove it?"

docs/notebooks/host_source_models.ipynb
jeremykubica: Done

docs/notebooks/citations.ipynb
jeremykubica: Yeah. We don't actually use the AGN model beyond extracting citations, but I agree it is better not to be inconsistent.

mi-dai (approved):

PR #897, opened 2026-07-06
Issue891

### LINKED ISSUE #891: JOSS review: Minor comments

jeremykubica:
Closes #891 

Add a description of the noise models.

### DIFF
--- a/.github/workflows/asv-main.yml
+++ b/.github/workflows/asv-main.yml
@@ -30,7 +30,7 @@ jobs:
       with:
         python-version: ${{env.PYTHON_VERSION}}
     - name: Checkout main branch of the repository
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
       with:
         fetch-depth: 0
     - name: Install dependencies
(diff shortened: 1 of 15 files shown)


hombit (commented):
Looks good, but what about `pyarrow>19.0.1`? It may be a nested-pandas problem, do we have a reproducer for a nested-pandas issue?

jeremykubica:
I made the pyarrow change (and some other fixes) in an earlier PR (#896)

hombit (approved):

PR #902, opened 2026-07-08
Add replay functionality to a simulation

jeremykubica:
All the simulation code to take a `GraphState` that overrides parameter sampling. This is useful if you want to run the exact same set of objects through two different survey designs, want to simulate the same objects with different noise, or want to save objects to replay later.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -103,7 +103,7 @@ in multiple observation tables and corresponding passband groups. See the
 Can I Rerun a Simulation with the Same Parameters?
 -------------------------------------------------------------------------------
 
-Yes. There are two approaches do doing this. 
+Yes. There are two approaches to doing this. 
 
 First, if you want to produce exactly the same results, you can provide a random number generator with a fixed seed to the ``simulate_lightcurves()`` function. This will ensure that the same random numbers are used in the simulation for both parameter sampling and noise generation, resulting in identical outputs.
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -100,6 +100,31 @@ in multiple observation tables and corresponding passband groups. See the
 :doc:`multiple surveys demo notebook <notebooks/multiple_surveys>` for an example of how to do this.
 
 
+Can I Rerun a Simulation with the Same Parameters?
+-------------------------------------------------------------------------------
+
+Yes. LightCurveLynx allows you to rerun a simulation with the same parameters by using the ``GraphState`` object. The ``GraphState`` object captures the state of the simulation graph and allows you to rerun the simulation with the same parameters. You can even change the survey information. This is particularly useful if you want to compare the results you get from different surveys on the exact same set of objects.
+
+You can capture the state of the previous simulation from the "params" column in its results table:
+
+.. code-block:: python
+
+    previous_state = GraphState.from_list(results["params"].values)
+
+Then you pass this ``previous_state`` to the ``simulate_lightcurves()`` function to rerun the simulation with the same parameters:
+
+.. code-block:: python
+
+    results2 = simulate_lightcurves(
+        model,
+        previous_state.num_samples,
+        new_survey_info,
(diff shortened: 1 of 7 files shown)


src/lightcurvelynx/simulate.py
```diff
@@ -190,6 +212,7 @@ def split(self, num_batches=None, batch_size=None):
                 output_file_path=batch_output_file_path,
                 save_full_filter_names=self.save_full_filter_names,
                 progress_bar=False,  # Don't show per-batch progress bars.
+                graph_state=batch_graph_state,
                 **self.kwargs,
             )
             batches.append(batch_info)
```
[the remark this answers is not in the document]
jeremykubica: Changes the constructor to not do a deep copy.

docs/faq.rst
```diff
@@ -100,6 +100,31 @@ in multiple observation tables and corresponding passband groups. See the
...
+Can I Rerun a Simulation with the Same Parameters?
+-------------------------------------------------------------------------------
+
+Yes. LightCurveLynx allows you to rerun a simulation with the same parameters by using the ``GraphState`` object. The ``GraphState`` object captures the state of the simulation graph and allows you to rerun the simulation with the same parameters. You can even change the survey information. This is particularly useful if you want to compare the results you get from different surveys on the exact same set of objects.
+
+You can capture the state of the previous simulation from the "params" column in its results table:
+
+.. code-block:: python
+
+    previous_state = GraphState.from_list(results["params"].values)
```
mi-dai, line 116: Can we overwrite some parameters? For example, t0, because the same object with same t0 may not be observable if one of the surveys was in the past but the other is in the future.

docs/faq.rst
```diff
@@ -100,6 +100,31 @@ in multiple observation tables and corresponding passband groups. See the
 :doc:`multiple surveys demo notebook <notebooks/multiple_surveys>` for an example of how to do this.
 
 
+Can I Rerun a Simulation with the Same Parameters?
+-------------------------------------------------------------------------------
+
+Yes. LightCurveLynx allows you to rerun a simulation with the same parameters by using the ``GraphState`` object. The ``GraphState`` object captures the state of the simulation graph and allows you to rerun the simulation with the same parameters. You can even change the survey information. This is particularly useful if you want to compare the results you get from different surveys on the exact same set of objects.
```
mi-dai: Maybe mention if you don't need to change survey info then simply set random number generator with the fix seed and it'll be the same?

docs/faq.rst
jeremykubica: Technically yes, but you need to be careful. If something depends on that variable, it will not automatically propagate to dependent parameters.

mi-dai: Would we rather suggesting the user to shift the obstable time column? Since we suggest the purpose is mainly to compare surveys, we should probably give some suggestions on how to do that.

docs/faq.rst
jeremykubica: I added this.

docs/faq.rst
jeremykubica: I added a few lines about this to the documentation.

docs/faq.rst
```diff
@@ -103,7 +103,11 @@ in multiple observation tables and corresponding passband groups. See the
 Can I Rerun a Simulation with the Same Parameters?
 -------------------------------------------------------------------------------
 
-Yes. LightCurveLynx allows you to rerun a simulation with the same parameters by using the ``GraphState`` object. The ``GraphState`` object captures the state of the simulation graph and allows you to rerun the simulation with the same parameters. You can even change the survey information. This is particularly useful if you want to compare the results you get from different surveys on the exact same set of objects.
+Yes. There are two approaches do doing this.
```
mi-dai: "to" doing

mi-dai (approved):

PR #908, opened 2026-07-14
Data-driven noise NB for DP1

hombit:
Adds an RSP pre-executed notebook with an example of data-driven noise generation using pzflow.

It also changes the pzflow noise model to clip bandfluxes to the minimum value, so we can apply the model to zero (or very small) "perfect" bandflux.


jeremykubica (approved):
This looks really good. Thank you for putting it together.

PR #925, opened 2026-08-05
Precompute Saturation as nJy once per ObsTable

jeremykubica:
Speed up the saturation computation by precomputing the conversion of the saturation thresholds in nJy. This is very small and doesn't really impact the running time much.

### DIFF
--- a/src/lightcurvelynx/obstable/obs_table.py
+++ b/src/lightcurvelynx/obstable/obs_table.py
@@ -82,8 +82,8 @@ class ObsTable:
         The saturation thresholds in magnitudes for each filter. If unspecified, an
         instrument-specific default will be used, if available.
     _saturation_njy : dict, optional
-        The saturation thresholds in nJy for each filter. If unspecified, an
-        instrument-specific default will be used, if available.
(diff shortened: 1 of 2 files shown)


olivialynn (approved):
Disclaimer that I'm taking your word for it that `np.hypot(flux_error, flux - expected_saturated_flux)` is the desired saturated flux error, but vibes-based the science seems right!

An aside that Kostya has long ago encouraged me to pre-calculate certain values and hardcode them into unit tests rather than testing an implementation against the same code path, but I don't know if this is the same code path, and regardless would probably be overkill here anyway.

jeremykubica:
> Disclaimer that I'm taking your word for it that `np.hypot(flux_error, flux - expected_saturated_flux)` is the desired saturated flux error, but vibes-based the science seems right!
> 
> An aside that Kostya has long ago encouraged me to pre-calculate certain values and hardcode them into unit tests rather than testing an implementation against the same code path, but I don't know if this is the same code path, and regardless would probably be overkill here anyway.

That's a fair point. I changed this to use a test that should always hold: the error after saturation is greater than or equal to the error before saturation.

PR #939, opened 2026-08-12
Update Spectrograph evaluation

jeremykubica:
Fix the spectrograph evaluation to correctly apply bin width (to change from flux densities to measured flux). Note that this changes the expected output of all previous tests using `evaluate`.

Also allow the spectrograph to sample at multiple points when the bins are large by providing a `max_wave_step` parameter that ensures we have a sample at least every `max_wave_step` within a bin. The code maintains short cut paths were the it just uses the center of the bins if no `max_wave_step` is provided or if the bins are already small enough.


src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -145,25 +237,8 @@ def from_midpoints(cls, wave_midpoints, bin_width, **kwargs):
...
-        float
-            The width of the bin in Angstroms.
-        """
-        if index < 0 or index >= len(self.waves_min):
-            raise IndexError(f"Index {index} out of bounds for {len(self.waves_min)} bin widths.")
-        return self.bin_widths[index]
-
     def wave_bounds(self):
-        """Get the minimum and maximum wavelength for this spectra.
+        """Get the minimum and maximum bin boundaries for this spectra.
```
mi-dai:
```suggestion
        """Get the minimum and maximum wavelength bin boundaries for this spectra.
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -10,73 +10,165 @@ class Spectrograph:
...
         The width of each wavelength bin in Angstroms.
+    all_bin_min : float
+        The minimum bin boundary of the spectra in Angstroms.
+    all_bin_max : float
+        The maximum bin boundary of the spectra in Angstroms.
+    waves : np.ndarray
+        The points at which to evaluate the flux density of the object in Angstroms. By default
+        this is the midpoint of each bin.
+    num_query_waves : int
+        The number of wavelength points at which to evaluate the flux density of the object.
```
mi-dai:
```suggestion
        The number of wavelength points at which to evaluate the flux density of the spectral model.
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -10,73 +10,165 @@ class Spectrograph:
...
+    num_bins : int
+        The number of bins in the spectra.
     bin_widths : np.ndarray
         The width of each wavelength bin in Angstroms.
+    all_bin_min : float
+        The minimum bin boundary of the spectra in Angstroms.
+    all_bin_max : float
+        The maximum bin boundary of the spectra in Angstroms.
+    waves : np.ndarray
+        The points at which to evaluate the flux density of the object in Angstroms. By default
```
mi-dai:
```suggestion
        The points at which to evaluate the flux density of the spectral model in Angstroms. By default
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -10,73 +10,165 @@ class Spectrograph:
...
     Attributes
     ----------
     waves_min : np.ndarray
         The start of each wavelength bin in Angstroms.
     waves_max : np.ndarray
         The end of each wavelength bin in Angstroms.
-    waves : np.ndarray
-        The center of each wavelength bin in Angstroms.
+    num_bins : int
+        The number of bins in the spectra.
```
mi-dai:
```suggestion
        The number of bins of the spectrograph.
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -10,73 +10,165 @@ class Spectrograph:
...
     bin_widths : np.ndarray
         The width of each wavelength bin in Angstroms.
+    all_bin_min : float
+        The minimum bin boundary of the spectra in Angstroms.
+    all_bin_max : float
+        The maximum bin boundary of the spectra in Angstroms.
+    waves : np.ndarray
+        The points at which to evaluate the flux density of the object in Angstroms. By default
+        this is the midpoint of each bin.
```
mi-dai:
```suggestion
        this is the midpoint of each bin (when `wave_step` is None or small enough).
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -10,73 +10,165 @@ class Spectrograph:
...
-        The center of each wavelength bin in Angstroms.
+    num_bins : int
+        The number of bins in the spectra.
     bin_widths : np.ndarray
         The width of each wavelength bin in Angstroms.
+    all_bin_min : float
+        The minimum bin boundary of the spectra in Angstroms.
+    all_bin_max : float
+        The maximum bin boundary of the spectra in Angstroms.
+    waves : np.ndarray
```
mi-dai:
```suggestion
    query_waves : np.ndarray
```

mi-dai: If I understand this correctly, this is the wavelengths to evaluate the model, and then being integrated over the bins. I suggest calling it `query_waves`. 
We may consider still keeping a `wave` attribute to record the actual wavelength for the spectrograph output.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -183,24 +258,67 @@ def evaluate(
         Parameters
         ----------
         flux_density_matrix : np.ndarray
-            A 2D or 3D array of flux densities. If the array is 2D it contains a single sample where
-            the rows are the T times and columns are M wavelengths in Angstroms. If the array is 3D
-            it contains S samples and the values are indexed as (sample_num, time, wavelength).
+            A 1D, 2D or 3D array of flux densities. The last dimension contains the flux density values
+            at the wavelengths specified by self.waves for a single sample.
```
mi-dai: What do the other dimensions represent?

tests/lightcurvelynx/astro_utils/test_spectrograph.py
```diff
@@ -154,7 +205,140 @@ def test_create_spectrograph_with_scale():
...
+        3566.7,  # Bin 0 - Sample 1
+        3625.0,  # Bin 1 - Sample 0
+        3675.0,  # Bin 2 - Sample 0
+        3733.3,  # Bin 3 - Sample 0
+        3766.7,  # Bin 3 - Sample 1
+        3840.0,  # Bin 4 - Sample 0
+        3880.0,  # Bin 4 - Sample 1
+        3920.0,  # Bin 4 - Sample 2
+        3960.0,  # Bin 4 - Sample 3
+    ]
```
mi-dai, line 380: I don't quite understand why we need to do it this way. I think for wave_step=50 I'm expecting [3500, 3550, 3600, 3650, ..., 3850, 3900, 3950, 4000]

mi-dai (commented):
I think we should assume non-overlapping bins and require waves_min and waves_max strictly increasing to make the code simpler. I don't see a use case that would need more flexibility

tests/lightcurvelynx/astro_utils/test_spectrograph.py
jeremykubica: I was debating between the two approaches. The fixed grid you describe would be simpler and faster to implement. However, it might not sample the bins evenly. Depending on the step size vs bin width, we might end up with: some bins sampled near the center, some bin sampled near the edges, and some bins sampled multiple times. This approach, which more complex, ensure that the samples within each bin are evenly spaced to allow us to directly average them. Let me know if you think this is overkill.

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: I added this to the comment

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: I changed this (and the code that uses it).

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -173,29 +253,73 @@ def evaluate(
...
+        # into one dimension) and each column corresponds to a wavelength.
+        # We do this so we can efficiently perform integration, scaling, etc. on all samples
+        # at once regardless of whether the input is 1D, 2D, or 3D.
+        initial_dimensions = flux_density_matrix.shape[:-1]
+        flux_density_flat = flux_density_matrix.reshape(-1, num_query_waves)
+
+        # Convert the flux density at each query wavelength into a flux for each query bin.
+        # We use rectangular interpolation, so the flux is just the flux density at the query
+        # wavelength (center of the bin) multiplied by the width of the query bin.
+        query_bin_flux_flat = flux_density_flat * self._query_widths[np.newaxis, :]
```
hmperk18, line 319: If we are expecting `flux_density_matrix` to have units of nJy, then we will need to convert Fnu to Flambda so Flux = F_lambda * deltaLambda.  This can by done by $F_{\lambda} = F_{\nu} * c / (\lambda^2)$, which I think would be something like`flux_density_matrix * astropy.const.c.to(u.AA/u.s).value / (self.query_waves**2)`

tests/lightcurvelynx/astro_utils/test_spectrograph.py
```diff
@@ -101,10 +112,17 @@ def test_spectrograph_equals():
...
-    """Test that we can create and query a Spectrograph object."""
+    """Test that we can create and query a Spectrograph object with a scale."""
     scale = np.array([0.5, 1.0, 1.0, 1.0, 0.8])
     spgraph = Spectrograph.from_regular_grid(wave_start=4000, wave_end=5000, bin_width=200.0, scale=scale)
     assert np.allclose(spgraph.query_waves, np.array([4100.0, 4300.0, 4500.0, 4700.0, 4900.0]))
+    assert spgraph.num_bins == 5
+
+    # One dimensional fluxes to spec_fluxes
+    measurement = np.array([50.0, 40.0, 20.0, 20.0, 10.0])
+    expected = np.array([25.0, 40.0, 20.0, 20.0, 8.0]) * spgraph.bin_widths
```
hmperk18: if we convert to Flambda as I mentioned, then all these expected values in the tests will be different

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -21,49 +26,114 @@ class Spectrograph:
...
+            if max_wave_step <= 0:
+                raise ValueError(f"max_wave_step must be positive, got {max_wave_step}.")
+
+            # For each bin: compute the number of points that need to be sampled and
+            # spread them evenly throughout the bin.
+            query_waves = []
+            wave_to_bin_map = []
+            query_widths = []  # The width of each query wavelength point (used for integration).
+            self._bin_counts = np.zeros(self.num_bins, dtype=int)
+            for bin_idx, (w_min, w_max) in enumerate(zip(self.waves_min, self.waves_max, strict=False)):
```
hombit, line 106: This may be "vectorized", but I guess it is not even close to be a bottle-neck

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -21,49 +26,114 @@ class Spectrograph:
...
+            self.query_waves = np.array(query_waves, dtype=float)
+            self._query_widths = np.array(query_widths, dtype=float)
+
+            # In the waves original order save the mapping from wave index to bin index and a mapping
+            # of bin index to where the bin starts in the waves array.
+            self._wave_to_bin_map = np.array(wave_to_bin_map, dtype=int)
+            self._bin_starts = np.concatenate(([0], np.cumsum(self._bin_counts)[:-1]))
+
+        # The query wavelengths should always be in increasing order.
+        if not np.all(np.diff(self.query_waves) > 0):  # pragma: no cover
```
hombit, line 127: Does this also cover L76?

src/lightcurvelynx/astro_utils/spectrograph.py
hombit: This is very suspicious, why do we integrate F_ν by λ?

hombit (approved):
LGTM. but I'm out of the context of the Spectrograph stuff, so may miss something.

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: I think this line is not technically needed because of line 76, but I want to leave it in to be clear.

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: Agreed in both that this may be able to be vectorized, but that it isn't a bottleneck right now. So I'm going to leave it as-is.

src/lightcurvelynx/astro_utils/spectrograph.py
mi-dai: I agree with @hmperk18

mi-dai (approved):

PR #941, opened 2026-08-13
Small Spectrograph changes

jeremykubica:
This PR pulls out some of the small and independent changes from #939 while we discuss the details of that PR. The goal is to submit the small changes now to simplify both discussion and a later review.

Changes include:
- Add a `num_bins` attribute.
- Rename `waves` to `query_waves` for clarity (suggested by @mi-dai)
- Add a check on non-positive bin widths
- Remove the `bin_width` function since we have the information stored in the `bin_widths` array.
- Rename `input` to `measurement` in the tests so the variable does not overlap with the builtin function (Copilot suggestion)
- Add tests
- Fix comments

### DIFF
--- a/src/lightcurvelynx/astro_utils/spectrograph.py
+++ b/src/lightcurvelynx/astro_utils/spectrograph.py
@@ -16,10 +16,14 @@ class Spectrograph:
         The start of each wavelength bin in Angstroms.
     waves_max : np.ndarray
         The end of each wavelength bin in Angstroms.
-    waves : np.ndarray
-        The center of each wavelength bin in Angstroms.
+    num_bins : int
+        The number of bins of the spectrograph.
     bin_widths : np.ndarray
         The width of each wavelength bin in Angstroms.
+    query_waves : np.ndarray
+        The points at which to evaluate the flux density of the spectral model in Angstroms.
+    num_query_waves : int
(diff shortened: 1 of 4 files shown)


src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -65,16 +73,20 @@ def __str__(self) -> str:
...
+        """Determine if two spectrographs have equal values for their internal data."""
+        if self.num_bins != other.num_bins:
             return False
         if not np.allclose(self.waves_min, other.waves_min):
             return False
         if not np.allclose(self.waves_max, other.waves_max):
             return False
+        if not np.allclose(self.bin_widths, other.bin_widths):  # pragma: no cover
+            return False
+        if self.instrument != other.instrument:  # pragma: no cover
```
olivialynn, line 88: How long are waves arrays expected to be? Might be slightly faster to put this higher to short circuit on the string comparison - but only if they'd get very long, probably inconsequential

olivialynn (approved):

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: While wavelengths will be longer than the strings, I'm guessing most of the instrument names will be left at their default.

PR #942, opened 2026-08-17
Support Redback model meta data

jeremykubica:
Redback recently added a `ModelMetadata` class that includes information on the phase bounds. Add code to support this class, using the metadata if the bounds are not explicitly provided.


src/lightcurvelynx/models/redback_models.py
```diff
@@ -67,6 +69,9 @@ class RedbackWrapperModel(SEDModel, CiteClass):
     time_extrapolation : tuple of Extrapolation, optional
         The extrapolation method to use for times outside the model's bounds.
         If nothing is provided, then the code adds zero padding.
+    model_metadata : redback.model_library.ModelMetadata, optional
+        The additional redback meta data. This can be accessed via redback's
+        get_model_metadata(<MODEL_NAME>) function.
```
hombit: Please tell here that we call it for the `None` value

hombit (approved):

src/lightcurvelynx/models/redback_models.py
jeremykubica: Fixed

PR #945, opened 2026-08-21
Code cleanups

jeremykubica:
Inspired by copilots suggestions in #944 , I manually checked and cleaned up some code throughout the project, including:

Efficiency:
- Change `np.array` to `np.asarray` where possible
- Change `rng = np.random.default_rng(rng)` to `if rng is None:`. While the former is more general (can take int seeds), our API requires rng to either be a random number generator or None. The `if` check is 1/4 the time when rng is already defined. They take the same time when rng is None.

Correctness:
- Add some more array dimensionality checks.
- Add some more checks to `validate_ra_dec_degrees`

### DIFF
--- a/tests/lightcurvelynx/obstable/test_obs_table.py
+++ b/tests/lightcurvelynx/obstable/test_obs_table.py
@@ -476,7 +476,7 @@ def test_obs_table_filter_rows():
     with pytest.raises(ValueError):
         _ = ops_data.filter_rows(bad_mask)
 
-    # We throw and error if the rows array is not one-dimensional.
+    # We throw an error if the rows array is not one-dimensional.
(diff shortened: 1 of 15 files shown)


tests/lightcurvelynx/obstable/test_obs_table.py
```diff
@@ -476,6 +476,16 @@ def test_obs_table_filter_rows():
     with pytest.raises(ValueError):
         _ = ops_data.filter_rows(bad_mask)
 
+    # We throw and error if the rows array is not one-dimensional.
```
olivialynn:
```suggestion
    # We throw an error if the rows array is not one-dimensional.
```

olivialynn (approved):

PR #946, opened 2026-08-21
Increase test coverage

jeremykubica:
Increase the test coverage through a combination of adding tests and marking some edge conditions as `# pragma: no cover`

In the process, I found the function `add_column_mapping` did not work as expected (and was untested). It has been removed.

### DIFF
--- a/tests/lightcurvelynx/test_graph_state.py
+++ b/tests/lightcurvelynx/test_graph_state.py
@@ -570,7 +570,7 @@ def test_graph_state_equal():
     state6 = GraphState(num_samples=2)
     assert state5 != state6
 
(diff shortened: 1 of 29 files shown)


tests/lightcurvelynx/test_graph_state.py
```diff
@@ -570,6 +570,22 @@ def test_graph_state_equal():
     state6 = GraphState(num_samples=2)
     assert state5 != state6
 
+    # Two states are not equal if they have sets of parameters.
```
olivialynn:
```suggestion
    # Two states are not equal if they have different sets of parameters.
```

olivialynn (approved):

PR #949, opened 2026-08-25
Update rubin_dp2.ipynb

jeremykubica:
Update the end to end Rubin DP2 notebook to use `rest_time_window_offset` to only generate "interesting" points for each supernova. This prevents the problem of generating all the points way before the supernovae peaks.


hombit (approved):
Very nice!

PR #951, opened 2026-08-27
Apply gaussian smearing to flux in spectrograph bins

hmperk18:
## Change Description
Following the SNANA implementation, applies gaussian smearing to flux into neighboring spectrograph bins based on provided wavelength resolutions per bin.  Pads the edges of the bins to allow flux from beyond the bins to possible smear into the specified bins.

## Code Quality
- [ ] I have read the Contribution Guide and agree to the Code of Conduct
- [ ] My code follows the code style of this project
- [ ] My code builds (or compiles) cleanly without any errors or warnings
- [ ] My code contains relevant comments and necessary documentation

### DIFF
--- a/src/lightcurvelynx/astro_utils/spectrograph.py
+++ b/src/lightcurvelynx/astro_utils/spectrograph.py
@@ -113,9 +113,7 @@ class Spectrograph:
         else:
             self.wavelength_resolution = np.asarray(wavelength_resolution, dtype=float)
         if len(self.wavelength_resolution) != self.num_bins:
-            raise ValueError(
-                "wavelength_resolution must have the same length as waves_min and waves_max."
-            )
+            raise ValueError("wavelength_resolution must have the same length as waves_min and waves_max.")
 
         if self.num_bins <= 0:  # pragma: no cover
             raise ValueError("Spectrograph must have at least one bin.")
--- a/src/lightcurvelynx/astro_utils/spectrograph.py
+++ b/src/lightcurvelynx/astro_utils/spectrograph.py
@@ -15,7 +15,6 @@ def gaussian_integral(nsigma_low, nsigma_high):
 
     Parameters
     ----------
-
     nsigma_low : float
         The lower limit of the integral in units of sigma.
     nsigma_high : float
@@ -94,8 +93,9 @@ class Spectrograph:
             The multiplicative factor to apply to each bin's flux. If None, no additional scaling
             is applied.
             Default: None
-        wavelength_resolution : np.ndarray
+        wavelength_resolution : np.ndarray | float
             The Gaussian sigma wavelength resolution for each bin in Angstroms.
+            If float, the same value is applied to all bins. If None, no smearing is applied.
         compute_smear : bool, optional
             Flag to enable smearing of flux between bins based on the wavelength resolution.
             If true, fluxes from Spectrograph.evaluate() will be smeared.
@@ -105,11 +105,17 @@ class Spectrograph:
         self.waves_min = np.asarray(waves_min, dtype=float)
         self.waves_max = np.asarray(waves_max, dtype=float)
         self.num_bins = len(self.waves_min)
-        self.wavelength_resolution = (
-            np.asarray(wavelength_resolution, dtype=float)
-            if wavelength_resolution is not None
-            else np.zeros(self.num_bins, dtype=float)
-        )
+
+        if wavelength_resolution is None:
+            self.wavelength_resolution = np.zeros(self.num_bins, dtype=float)
+        elif np.isscalar(wavelength_resolution):
+            self.wavelength_resolution = np.full(self.num_bins, float(wavelength_resolution), dtype=float)
+        else:
+            self.wavelength_resolution = np.asarray(wavelength_resolution, dtype=float)
+        if len(self.wavelength_resolution) != self.num_bins:
+            raise ValueError(
+                "wavelength_resolution must have the same length as waves_min and waves_max."
+            )
 
         if self.num_bins <= 0:  # pragma: no cover
             raise ValueError("Spectrograph must have at least one bin.")
@@ -244,7 +250,7 @@ class Spectrograph:
 
     def _compute_padded_bins(self):
         """Compute the parameters for padded bins on either side of the main spectrograph bins. This allows
-        for flux beyond the edges of the spectrograph to be possibily smeared into the main bins.
+        for flux beyond the edges of the spectrograph to be possibly smeared into the main bins.
 
         Returns
         -------
@@ -261,6 +267,9 @@ class Spectrograph:
 
         # determine the extended region
         # snana expanded by 2.5 sigma of the edge bins
+        # note: if the wavelength resolution negative for a given side, then the num_pad_[side] is 0,
+        # the pad_[side]_min/max are empty, and no padding is added on that side.
+        # otherwise, at least 1 bin will be added on a given side
         sigma_blue = self.wavelength_resolution[0]
         width_blue = self.bin_widths[0]
         num_pad_blue = int(2.5 * sigma_blue / width_blue) + 1 if sigma_blue > 0 else 0
--- a/src/lightcurvelynx/astro_utils/spectrograph.py
+++ b/src/lightcurvelynx/astro_utils/spectrograph.py
@@ -229,15 +229,17 @@ class Spectrograph:
         return True
 
     def _compute_padded_bins(self):
-        """Compute the parameters for padded bins on either side of the main spectrograph bins. This allows 
+        """Compute the parameters for padded bins on either side of the main spectrograph bins. This allows
         for flux beyond the edges of the spectrograph to be possibily smeared into the main bins.
 
         Returns
         -------
         padded_bins_min : np.ndarray
-            the minimum wavelength of each padded bin in Angstroms.
+            the minimum wavelength of each bin, include the new padding bins, in Angstroms.
         padded_bins_max : np.ndarray
-            the maximum wavelength of each padded bin in Angstroms.
+            the maximum wavelength of each bin, include the new padding bins, in Angstroms.
+        padded_wavelength_resolution : np.ndarray
+            the wavelength resolution of each bin, include the new padding bins, in Angstroms.
         is_padding : np.ndarray
             a boolean array indicating which bins are padding (True) and which are in the
             original spectrograph.
@@ -263,11 +265,13 @@ class Spectrograph:
         padded_bins_min = np.concatenate((pad_blue_min, self.waves_min, pad_red_min))
         padded_bins_max = np.concatenate((pad_blue_max, self.waves_max, pad_red_max))
 
-        padded_wavelength_resolution = np.concatenate([ # TODO: needed?
-            np.full(num_pad_blue, sigma_blue),
-            self.wavelength_resolution,
-            np.full(num_pad_red, sigma_red)
-        ])
+        padded_wavelength_resolution = np.concatenate(
+            [
+                np.full(num_pad_blue, sigma_blue),
+                self.wavelength_resolution,
+                np.full(num_pad_red, sigma_red)
+            ]
+        )
 
         is_padding = np.concatenate(
             [
--- a/src/lightcurvelynx/astro_utils/spectrograph.py
+++ b/src/lightcurvelynx/astro_utils/spectrograph.py
@@ -8,6 +8,7 @@ from astropy import units as u
 from lightcurvelynx.astro_utils.unit_utils import fnu_to_flam
 import scipy
 
+
 def gaussian_integral(nsigma_low, nsigma_high):
     """
     Computes the integral of a Gaussian between two limits in units of sigma.
@@ -25,7 +26,10 @@ def gaussian_integral(nsigma_low, nsigma_high):
     integral : float
         The integral of the Gaussian between the limits.
     """
-    return 0.5 * (scipy.special.erf(nsigma_high / np.sqrt(2.0)) - scipy.special.erf(nsigma_low / np.sqrt(2.0)))
+    return 0.5 * (
+        scipy.special.erf(nsigma_high / np.sqrt(2.0)) - scipy.special.erf(nsigma_low / np.sqrt(2.0))
+    )
+
 
 class Spectrograph:
     """Models all of the bins of a spectrograph, producing bandfluxes for each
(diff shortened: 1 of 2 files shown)


src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -6,7 +6,13 @@
 from astropy import units as u
 
 from lightcurvelynx.astro_utils.unit_utils import fnu_to_flam
+import scipy
 
+def gaussian_integral(nsig_low, nsig_high):
+    """
+    Compute the integral of a Gaussian between two limits in units of sigma.
```
jeremykubica: To be consistent with the documentation in the rest of the code package, can you add a Parameters and Returns section?

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -63,15 +73,26 @@ def __init__(
             that are evaluated while computing the bin's flux. The smaller this value, the more
             accurate and expensive the integration. If None, a single sample per bin is used.
             Default: None
+        oversample_factor: int, optional
```
jeremykubica: As mentioned in slack, I think we should only have one of either `oversample_factor`and `max_wave_step`.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -63,15 +73,26 @@ def __init__(
...
+            If max_wave_step is not provided, divide each bin into this many sub-bins to evaluate
```
jeremykubica: The comment for this parameter is for `max_wave_step`.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -63,15 +73,26 @@ def __init__(
...
             is applied.
             Default: None
+        wavelength_resolution : np.ndarray
+            The Gaussian sigma wavelength resolution for each bin in Angstroms.
         """
         # Check that the input arrays are valid and convert them to numpy arrays.
         self.waves_min = np.asarray(waves_min, dtype=float)
         self.waves_max = np.asarray(waves_max, dtype=float)
         self.num_bins = len(self.waves_min)
+
```
jeremykubica, line 117: If we do keep both `oversample_factor` and `max_wave_step`, it would probably be easiest just to check that only one is set. Something like:
```
if oversample_factor is not None and max_wave_step is not None:
    raise ValueError("Only one of oversample_factor and max_wave_step can be set at a time.")
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -63,15 +73,26 @@ def __init__(
...
+        oversample_factor = 1 if wavelength_resolution is None else oversample_factor
```
jeremykubica: It's not clear to me what you would want to set oversample factor to 1 if there is no `wavelength_resolution` given. It seems like there could still be times you'd still want to over sample the bins.

tests/lightcurvelynx/astro_utils/test_spectrograph.py
```diff
@@ -200,18 +200,36 @@ def test_create_spectrograph_from_bad_bins():
         _ = Spectrograph(waves_min_ooo, waves_max_ooo)
 
 
+def test_create_spectrograph_from_bad_bins():
```
jeremykubica, line 190: This repeats the test name from another test, so only one of them will be run.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
         if self.scale is not None or other.scale is not None:
             if self.scale is None or other.scale is None:
                 return False
             if not np.allclose(self.scale, other.scale):
                 return False
         return True
+    
+
+    def _compute_padded_bins(self):
+        """Compute the padded bins for the spectrograph.
```
jeremykubica: These are bins before and after the main spectrograph bins, correct? If so, can you add that?

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+
+        Parameters
+        ----------
+        oversampling : int, optional
```
jeremykubica: `oversampling` isn't included as a method parameter. You can remove this block of the comment.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+            The factor by which to oversample the bins. Default is 10.
+
+        Returns
+        -------
+        padded_bins : np.ndarray
+            A 2D array of shape (num_bins, oversampling) representing the padded bins.
+        """
+
+        # determine the extended region
+            # snana expanded by 2.5 sigma of the edge bins
```
jeremykubica: Comment indented too many spaces.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+        Returns
+        -------
+        smear_matrix : np.ndarray
+            A 2D array of shape (num_bins, num_bins) representing the smearing matrix.
+            where smear_matrix[i, j] represents the fraction of flux from bin i that smears into j
+        """
+
+        smear_matrix = np.zeros((self.num_padded_bins, self.num_padded_bins))
+        for i in range(self.num_padded_bins):
+            sigma = self.padded_resolution[i]
```
jeremykubica, line 322: I think we can drop the `padded_resolution` attribute and just do:

```
sigma = self.padded_max[i] - self.padded_min[i]
```

Or, as a slight more efficient approach, you could do a vectorized compute of that before the loop:
```
sigma = self.padded_max - self.padded_min
```
and just reference `sigma[i]` within the loop.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -150,6 +194,8 @@ def __init__(
             flam_unit=u.erg / u.s / u.cm**2 / u.AA,
             fnu_unit=u.nJy,
         )
+        # compute the smear matrix, if there are wavelength resolutions
```
jeremykubica: I think we should also add a boolean parameter `compute_smear` that can shut this off even if `wavelength_resolution` is provided. That will allow uses to test on SNANA files without always using smear.

Maybe something like:
```
if compute_smear:
    self.wavelength_resolution is None:
        raise ValueError("wavelength_resolution required for smear computation.")
    self.smear_matrix = self._compute_smear_matrix()
else:
    self.smear_matrix = None
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+        smear_matrix : np.ndarray
+            A 2D array of shape (num_bins, num_bins) representing the smearing matrix.
+            where smear_matrix[i, j] represents the fraction of flux from bin i that smears into j
+        """
+
+        smear_matrix = np.zeros((self.num_padded_bins, self.num_padded_bins))
+        for i in range(self.num_padded_bins):
+            sigma = self.padded_resolution[i]
+
+            # Determine the range of the smearing
```
jeremykubica, line 324: I think this is making an assumption (which I'm guessing SNANA makes as well) that there are not significant gaps between the bins. If that is the case, then I would start the function with a test:

```
if np.any(np.abs(self.padded_max[:-1] - self.padded_mmin[1:]) > 1e-8)):
    raise ValueError("The spectrograph bins cannot have gaps when using smearing.") 
```

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+            lambda_bin = self.padded_widths[i]
+            n_bins_index = int(n_sigma * sigma / lambda_bin + 0.5)
+            j_low = max(i - n_bins_index, 0)
+            j_high = min(i + n_bins_index, self.num_padded_bins - 1)
+            bin_center = (self.padded_min[i] + self.padded_max[i]) / 2
+
+            # get smearing factor, if the bin is not padding
+            for j in range(j_low, j_high + 1):
+                # skip if in the padded region
+                if self.is_padding[j]:
```
jeremykubica, line 335: Should this be checking one loop up (skipping if the `i` bin is padding)?

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,12 +222,111 @@ def __eq__(self, other) -> bool:
...
+                    # compute the limits of the Gaussian integral in units of sigma
+                    lam_sig0 = (self.padded_min[j] - bin_center) / sigma
+                    lam_sig1 = (self.padded_max[j] - bin_center) / sigma
+                    smear_matrix[i, j] = gaussian_integral(lam_sig0, lam_sig1)
+                else: # sigma == 0, no smearing to outer bins
+                    smear_matrix[i, j] = 0.0 if i != j else 1.0
+
+        return smear_matrix
+
+    @property
```
jeremykubica, line 348: I already added this function in another PR, so you can remove this instancte.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -269,6 +414,7 @@ def wave_bounds(self):
     def evaluate(
         self,
         flux_density_matrix: np.ndarray,
+        smear: bool = True,
```
jeremykubica: As noted earlier in the review, I'm inclined to make smearing either always on or always off for the spectrograph. This will be more consistent with the way the spectrograph's interface is used. It will mean for testing, the user will need to create two spectrograph objects. But I think that is okay.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -331,18 +480,22 @@ def evaluate(
...
         # smearing matrix in the __init__ method and then apply it here. This will allow us to model
         # the effects of the spectrograph's point spread function on the measured fluxes.
-        # For now, we will skip this step.
+        if smear and self.smear_matrix is not None:
+            spectro_bin_flux @= self.smear_matrix
+
+        # get the flux in the spectrograph's native bins (not the padded bins)
+        spectro_bin_flux = spectro_bin_flux[..., ~self.is_padding]
+
+        # Multiply by any per-bin scaling factors.
```
jeremykubica, line 510: Questioning my code choices. Are there instances we will want to use scaling? I had been thinking that maybe different bins would have different sensitivities, but I didn't see anything like that in the SNANA code.

jeremykubica (changes requested):

src/lightcurvelynx/astro_utils/spectrograph.py
hmperk18: I think it might be necessary.  If we change the resolution of the padded bins to be the bin width, in the case where the bin width is larger than the resolution currently used, then the padded bins would smear flux further into main set of spectrograph bins.  Or the in the opposite case, not enough flux would be.
Would it be better to restructure the code such that the padded bin parameters aren't attributes and instead of temporary values?

src/lightcurvelynx/astro_utils/spectrograph.py
hmperk18: bin `i` is the source and smears into all `j` bins, and this check just prevents putting flux into the padded bins.  technically, I think they could receive flux, but in the end of evaluate, we return the flux in only the original bins, so we just skip adding the flux there.

jeremykubica: That makes sense. I was reading it backwards as `i` was receiving the flux.

src/lightcurvelynx/astro_utils/spectrograph.py
hmperk18: I think this sort of sensitivity comes from the spectrograph files, which come from the source spectrograph's ETC outputs, and is a combination of the wavelength resolution per bin and the signal-to-noise that is needed per bin, per reference magnitude, and per reference exposure time. Snana breaks this computation into two parts: smearing (just uses the wavelength resolution) and the poisson noise calculation (uses the full spectrograph file data)
All that is to say, I think it is okay to keep it in the case where a user would want a simple way of testing out this sort of efficiency per bin that the current scale is

jeremykubica: Sounds good. Thanks!

src/lightcurvelynx/astro_utils/spectrograph.py
jeremykubica: We can leave it as is.

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -3,11 +3,34 @@
...
 from lightcurvelynx.astro_utils.unit_utils import fnu_to_flam
 
 
+def gaussian_integral(nsigma_low, nsigma_high):
+    """
+    Computes the integral of a Gaussian between two limits in units of sigma.
+
+    Parameters
+    ----------
+
```
jeremykubica: Can you remove the blank line after the heading?

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -67,11 +94,23 @@ def __init__(
...
+        compute_smear : bool, optional
+            Flag to enable smearing of flux between bins based on the wavelength resolution.
+            If true, fluxes from Spectrograph.evaluate() will be smeared.
+            Default: False
         """
         # Check that the input arrays are valid and convert them to numpy arrays.
         self.waves_min = np.asarray(waves_min, dtype=float)
         self.waves_max = np.asarray(waves_max, dtype=float)
         self.num_bins = len(self.waves_min)
+        self.wavelength_resolution = (
```
jeremykubica: Do we need to add a check that the length of `wavelength_resolution` is the same as `waves_min`?

src/lightcurvelynx/astro_utils/spectrograph.py
```diff
@@ -176,13 +233,111 @@ def __eq__(self, other) -> bool:
...
+        pad_blue_max = pad_blue_min + width_blue
+
+        red_edge_idx = -2 if self.num_bins >= 2 else -1  # snana takes the 2nd to last bin
+        sigma_red = self.wavelength_resolution[red_edge_idx]
+        width_red = self.bin_widths[red_edge_idx]
+        num_pad_red = int(2.5 * sigma_red / width_red) + 1 if sigma_red > 0 else 0
+        pad_red_min = self.waves_max[-1] + width_red * np.arange(num_pad_red)
+        pad_red_max = pad_red_min + width_red
+
+        # combine the original bins with the padded bins
```
jeremykubica, line 284: It's worth adding a note that if either `num_pad_red` or `num_pad_blue` are zero, the associated min and max arrays will be empty. I had to work through a few examples to convince myself that is the case. So it would help future people reading the code.

jeremykubica (approved):

Issue #955, opened 2026-09-01
Three reproducible bugs found while validating against SNANA: add_effect() parameter collision, seed= dropped by internal samplers, SED wavelength range not checked


mauriciovalenzuela742:
## Context

We're running a long validation project comparing LightCurveLynx against SNANA (the reference C simulator used in production for an LSST/Rubin-adjacent pipeline), simulating ~14 real transient classes end-to-end and cross-checking against real SNANA `.DUMP`/`_HEAD.FITS`/`_PHOT.FITS` output on the same real cadence. Along the way we found three independent, reproducible bugs in the library. Filing them together since they surfaced from the same investigation, but they're unrelated to each other and touch different modules — happy to split into separate issues if that's preferred, or to share our reproduction scripts / open PRs for any of them.

Environment for all three: `lightcurvelynx` installed via `pip install "lightcurvelynx[all]"` (version 0.5.2, patterns also confirmed present on `main`), Python 3.12.3, Linux.

---

## 1. `PhysicalModel.add_effect()` silently drops a parameter when two effects register the same name

Breaks composing two `ExtinctionEffect` instances (e.g. Milky Way + host galaxy extinction).

`add_effect()` (`lightcurvelynx/models/physical_model.py`, around line 470-476) registers each effect's parameters by name:

```python
for param_name, setter in effect.parameters.items():
    if param_name not in self.setters:
        self.add_parameter(param_name, setter, ...)
```

If a second effect declares a parameter whose name is already registered by a previously-added effect, it is simply skipped — no exception, no warning, nothing logged. The second effect's own setter is never registered; at evaluation time it silently receives whatever value the first effect's setter produces instead.

This is not a hypothetical collision: `ExtinctionEffect.__init__()` (`lightcurvelynx/effects/extinction.py`), the library's own base class for extinction, always registers its parameter under the fixed name `"ebv"`, regardless of what the effect instance is conceptually for. Any model that adds two `ExtinctionEffect` instances — e.g. one for Milky Way foreground extinction and one for host galaxy extinction, the standard setup for any realistic supernova/transient simulation — hits this collision by construction. The second one added (in our case, host extinction) never gets its own setter registered, and receives the first one's (MW's) value at every evaluation.

[...]

PR #957, opened 2026-09-01
Fix part 1 of issue 955

jeremykubica:
## Change Description
Addresses the first part of #955 where `add_effect` is silently dropping parameters. This PR adds a check that the parameter name does not exist already OR that the new setter value is `None` (which means use the existing value). This later condition is needed to allow effects to access existing model parameters such as ra and dec.

Since this was discovered when chaining extinction models, we also add a path for extinction models to rename their ebv parameter. This allows users to apply multiple extinction models as long as they explicitly override the parameter names.


hombit (approved):
Could it also collide with a parameter introduced by a model?

jeremykubica:
> Could it also collide with a parameter introduced by a model?

Yes. This fixes that case too.

PR #959, opened 2026-09-01
Fix part 3 of issue 955

jeremykubica:
Addresses the third part of #955 where the `SEDTemplate.evaluate_sed()` method returns the last valid flux value for all wavelengths outside the SED's native wavelength range. This PR changes it to use zero instead (and updates the documentation).

In the process of fixing this, I found a bunch of other small fixes. Most significantly is that the lack of `minphase`, `maxphase`, `minwave`, and `maxwave` methods meant that the out of bounds was not causing a warning or using extrapolation.

Changes:
- `evaluate_sed` returns 0 outside of range
- `SEDTemplate` implements `minphase`, `maxphase`, `minwave`, and `maxwave` so it produces warnings and uses extrapolation.
- Fix the corresponding `minphase`, etc. implementations in the `RandomMultiObjectModel `minphase`, etc. implementations for the `AdditiveMultiObjectModel` as the sub-models are evaluated independently and the approach of returning a list wasn't needed.
- Baselines were not being interpolated for actual query wavelengths in `SEDTemplate`. Fix that.

### DIFF
--- a/src/lightcurvelynx/models/lightcurve_template_model.py
+++ b/src/lightcurvelynx/models/lightcurve_template_model.py
@@ -946,7 +946,7 @@ class MultiLightcurveTemplateModel(BaseLightcurveBandTemplateModel):
             raise ValueError("Filter must be provided to compute minphase.")
         if graph_state is None:
             raise ValueError("Graph state must be provided to compute minphase.")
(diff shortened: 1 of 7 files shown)


hombit (approved):
Looks good!

PR #963, opened 2026-09-09
Add built-in support for LSST detector footprint

jeremykubica:
Creates helper methods to generate a `DetectorFootprint` from a sorcha-format corners file and a helper method to generate a few known footprint types (including downloading the full corners file from the sorcha repository).  The PR also expands the documentation to make it easier for users to find information on how to use LSST footprints.


hombit (approved):
Corner file shouldn't be so large, should we just vendor it?

PR #965, opened 2026-09-10
Break out the evaluation of a single model on a single survey

jeremykubica:
Break out the core evaluation step of computing bandfluxes for a single sample of a single survey. This allows the possibility of calling just the evaluation from other code (specifically Rubin's metric analysis framework). The functionality of the current workflow is unchanged.


jeremykubica:
After discussion with @mi-dai and @hombit, closing this for now. Will figure out alternate approaches to doing this in the future.

PR #968, opened 2026-09-10
Switch Pandas .values to .to_numpy()

jeremykubica:
The values attribute is not recommended.

This is mostly a find-replace with manual editing of cases where we could move as `astype` into the function.


src/lightcurvelynx/obstable/roman_obstable.py
hombit, line 1: Why not .at[0] or something similar here?

hombit (approved):

src/lightcurvelynx/obstable/roman_obstable.py
jeremykubica: Because they are still in the Pandas frame at this time, they are retaining their original index and might not be at 0.  I found that out by converting it and running into an `KeyError` on the tests.

PR #974, opened 2026-09-20
Add helper function and fix fake SNANA data

jeremykubica:
Add a helper function for creating a `Spectrograph` directly from the snana file. In the process of testing this, I found a small error in the fake SNANA testing data, so I'm fixing that as well.


hmperk18 (approved):
looks good to me!

PR #975, opened 2026-09-21
Adding support for various blend types for pyLIMA_models wrapper

KKruszynska:
## Solution Description
I added support for `ftotal`,  `gblend`, and `noblend` for simulating microlensing events in the `pylima_models` wrapper. In principle, this could also encourage the user to pass source and blend magnitudes in different formats, but I did not change that. 

So:
- the user still passes `source_mags` and `blend_mags`;
- depending on the `blend_flux_parameter`, the source and blend magnitudes are turned into fluxes and the relevant `blend_flux_parameter`;
- I changed the go-to `blend_flux_parameter` to `ftotal`, because it tends to be more stable when fitting microlensing events (but it might not be important for simulating them).

One issue: 
- I keep getting errors from ruff on sections of the code that I did not modify when doing a pre-commit.

## Code Quality
- [x] I have read the Contribution Guide and agree to the Code of Conduct
- [x] My code follows the code style of this project
- [x] My code builds (or compiles) cleanly without any errors or warnings
- [x] My code contains relevant comments and necessary documentation


tests/lightcurvelynx/models/test_pylima_models.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 import pytest
+
```
jeremykubica: The linter is failing on this extra blank line.

tests/lightcurvelynx/models/test_pylima_models.py
```diff
@@ -84,3 +85,51 @@ def test_pylima_fail_create():
...
+    blend_mags = {
+        "g": 24.5,
+        "r": 24.0,
+    }
+    pylima_params = {
+        "u0": 0.01,
+        "tE": 25.0,
+    }
+
+    for blend_flux_parameter in ["ftotal", "gblend", "fblend"]:
```
jeremykubica, line 108: Not required here, but passing along for future consideration.

If you wanted, you could add:

```
with subtests.test(blend_flux_parameter=blend_flux_parameter):
```

and change the function signature to
```
def test_pylima_blends(subtests):
```

which would effectively create a different test for each value of blend_flux_parameter.

jeremykubica (commented):
One lint error to fix and the then one FYI.

tests/lightcurvelynx/models/test_pylima_models.py
KKruszynska: I just pushed an updated version and it doesn't complain anymore it seems.

tests/lightcurvelynx/models/test_pylima_models.py
KKruszynska: I added the requested changes in the latest push.

jeremykubica (approved):
Looks good. Feel free to merge it in.

Issue #511, opened 2025-09-26
Integration with MOSFIT


jeremykubica:
Allow LCL to use MOSFIT models, either by direct integration with MOSFIT or by reading their output models.

jeremykubica:
Closed by #570

PR #977, opened 2026-09-21
Add a lightweight MOSFiT wrapper model

### LINKED ISSUE #976: Wrap MOSFiT models as a LightCurveLynx SEDModel source
## Summary

Add a lightweight wrapper that lets LightCurveLynx use [MOSFiT](https://github.com/guillochon/MOSFiT) models as `SEDModel` sources, evaluated in process.

## Motivation

MOSFiT carries a large library of semi-analytic transient engines (SLSNe, magnetar-powered SNe, kilonovae, TDEs, CSM interaction, ...) with calibrated priors for each parameter. Those engines are exactly the kind of physical source LightCurveLynx wants to put behind a cadence and a noise model, but MOSFiT's own pipeline runs all the way to observed photometry — it redshifts the SED, applies line-of-sight extinction and integrates through bandpasses — which is the opposite of what a forward-simulation framework needs.

[guillochon/MOSFiT#252](https://github.com/guillochon/MOSFiT/pull/252) adds the missing view: a `mosfit.lynx.LynxSource` class whose `compute_sed` returns a `(n_phase, n_wave)` array of rest-frame flux densities in nJy at 10 pc, on a caller-supplied wavelength grid, with redshift, time dilation and extinction all pinned off so LightCurveLynx owns them. That makes a thin wrapper possible: LightCurveLynx supplies redshift, distance, dust and cadence, and MOSFiT supplies only the rest-frame SED.

This is distinct from #511, which was closed by #570 (`SIMSEDModel`, reading pre-generated SED grids off disk). This issue is about evaluating a MOSFiT model live, so its physical parameters can be sampled per object by the parameter graph.

## Proposed work

* New `src/lightcurvelynx/models/mosfit_models.py` with a `MOSFiTWrapperModel(SEDModel, CiteClass)`.
  * MOSFiT physical parameters (`mejecta`, `pspin`, `bfield`, ...) passed as a `parameters` dict become settable LightCurveLynx parameters, so they can be driven by any `ParameterizedNode`. Unspecified free parameters fall back to their MOSFiT prior midpoint.
  * `compute_sed` evaluates on a fixed internal phase/wavelength grid and interpolates onto the requested points. `LynxSource` rebuilds its whole MOSFiT stack when the grid changes, so the wrapper must hold one grid and vary only the parameters.
  * Output at 10 pc is rescaled by `(10 / distance_pc) ** 2`, matching `SIMSEDModel`. Redshift is left to LightCurveLynx (`apply_redshift` stays on), unlike `RedbackWrapperModel`.
  * `minwave`/`maxwave`/`minphase`/`maxphase` report the grid bounds so the existing extrapolation machinery engages, with `ZeroPadding` as the time default.
* Tests in `tests/lightcurvelynx/models/test_mosfit_models.py`, using an injectable toy source so the bulk of the suite runs without MOSFiT installed, plus `pytest.importorskip("mosfit")` for the real end-to-end path.
* `mosfit` stays an optional dependency, imported lazily with an install hint on `ImportError`.
* Docs: add MOSFiT to the wrapped-package lists in `index.rst` and `simulations.rst`.

## Dependency

Requires [guillochon/MOSFiT#252](https://github.com/guillochon/MOSFiT/pull/252) (MOSFiT 2.1) to be merged and released before the wrapper can be exercised against a real MOSFiT install.

guillochon:
Closes #976.

> [!IMPORTANT]
> **Do not merge until [guillochon/MOSFiT#252](https://github.com/guillochon/MOSFiT/pull/252) is merged and MOSFiT 2.1 is released.** This wrapper builds on `mosfit.lynx`, which that PR adds. Everything here is inert without it: the import is lazy, and the one end-to-end test is gated behind `pytest.importorskip("mosfit.lynx")`, so CI passes either way.

## Why

MOSFiT carries a large library of semi-analytic transient engines — SLSNe, magnetar-powered SNe, kilonovae, TDEs, CSM interaction — each with calibrated priors. Those are exactly the sources LightCurveLynx wants to put behind a cadence and a noise model. The obstacle has been that MOSFiT's own pipeline takes a model all the way to observed photometry: it redshifts the SED, applies line-of-sight extinction, integrates through bandpasses and compares against data. A forward-simulation framework wants the opposite.

[guillochon/MOSFiT#252](https://github.com/guillochon/MOSFiT/pull/252) adds the missing view: `mosfit.lynx.LynxSource.compute_sed` returns a rest-frame SED in nJy at 10 pc on a caller-supplied wavelength grid, with redshift, time dilation, explosion time and extinction all pinned so the calling simulator owns them. This PR is the thin wrapper on that side.

This is distinct from #511 / #570 (`SIMSEDModel`), which reads pre-generated SED grids off disk. Here the MOSFiT model is evaluated live, so its physical parameters can be sampled per object by the parameter graph.

## What's in it

`MOSFiTWrapperModel(SEDModel, CiteClass)` in `src/lightcurvelynx/models/mosfit_models.py`:

```python
model = MOSFiTWrapperModel(
    "slsn",
    parameters={"mejecta": mej_node, "vejecta": 1.0e4, "Pspin": 2.0},
    redshift=0.3,
    cosmology="Planck18",
    t0=t0_node,
)
```

- **MOSFiT parameters become LightCurveLynx parameters.** Anything named in `parameters` gets a setter and can be driven by any `ParameterizedNode`. Free parameters that are *not* named fall back to the midpoint of their MOSFiT prior, so a simulation can vary three parameters without having to name the other six. The names are MOSFiT's own and are case sensitive (`slsn` takes `Pspin` in ms, `Bfield` in 10^14 G, `Mns`, `thetaPB`, `mejecta` in solar masses, `kappa`, `kappagamma`, `vejecta` in km/s and `temperature` in K); an unrecognized name raises an error listing the model's free parameters.
- **LightCurveLynx owns the redshift**, unlike `RedbackWrapperModel`. MOSFiT hands back a rest-frame SED, so `apply_redshift` stays on and the superclass does the time dilation and wavelength shift. Dust goes on as an ordinary `EffectModel`.
- **Distance.** MOSFiT's output is at 10 pc, so the wrapper rescales by `(10 / distance_pc) ** 2`, matching `SIMSEDModel`. A model with no way to determine a luminosity distance is rejected in the constructor rather than at evaluation time.
- **One grid, held fixed.** `LynxSource` re-runs MOSFiT's whole data-loading path whenever its grid changes, so the wrapper evaluates on a single phase/wavelength grid — which it requests explicitly on every call — and interpolates onto the queried points. `minphase`/`maxphase`/`minwave`/`maxwave` report that grid so the existing extrapolation machinery engages, with `ZeroPadding` as the time default.
- **Caching.** `compute_sed` is called more than once per sample when extrapolation is in play, and a MOSFiT evaluation is far more expensive than a parameter comparison, so the interpolator for the last parameter set is cached.
- **MOSFiT stays optional.** The import is deferred to first evaluation and raises with an install hint. `pyproject.toml` carries a commented-out `mosfit>=2.1` entry in `dev` and `all`, to be enabled once 2.1 is on PyPI.

## Testing

`tests/lightcurvelynx/models/test_mosfit_models.py` — 13 tests. Most run against an injectable toy source, so they exercise the wrapper without MOSFiT installed: phase alignment against `t0`, distance scaling, behaviour outside the grid in both time and wavelength, caching, the redshift path, that the source is asked for the wrapper's own grid, and the constructor and evaluation error paths.

`test_mosfit_wrapper_real_source` builds a real MOSFiT `slsn` and evaluates it. It is skipped when `mosfit.lynx` is unavailable, and runs from a `tmp_path` because MOSFiT copies its `modules` directory into the working directory.

Verified locally against MOSFiT installed from the `claude/lynx-sed-api` branch — all 13 pass, including the end-to-end test. As a units sanity check, the prior-midpoint `slsn` peaks at a monochromatic M_AB ≈ −20.6 at 3500 Å, 15 days after explosion, which is the right neighbourhood for an SLSN. Instrumenting `LynxSource._build` confirms the explicit grid request costs no rebuild: one build across two evaluations with different parameters.

`pytest tests/lightcurvelynx/models` is otherwise unchanged (the only failures on my machine are the pre-existing `test_salt2.py` ones from JAX not being installed). `ruff check` and `ruff format` are clean at the pinned v0.2.1.

## Not in this PR

A demo notebook, to keep the diff small. Happy to add one under `docs/notebooks/pre_executed/` alongside the redback and pyLIMA examples if you'd like it here rather than as a follow-up — it would need MOSFiT 2.1 on PyPI first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


guillochon:
@mcoughlin would you be willing to review this one? It wraps MOSFiT's transient models as a LightCurveLynx `SEDModel`, on top of the rest-frame SED interface added in guillochon/MOSFiT#252.

It is a draft on purpose: it should not merge until that MOSFiT PR is merged and MOSFiT 2.1 is released. Everything here is inert without it, so it can be reviewed now and flipped to ready once the dependency lands.

I do not have permission to add reviewers on this repo from a fork, so flagging it here instead — could someone with write access add @mcoughlin as a formal reviewer?

src/lightcurvelynx/models/mosfit_models.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # Reuse the last result if the parameters have not changed. compute_sed() is called
+        # more than once per sample when extrapolation is in play, and a MOSFiT evaluation
+        # is far more expensive than the comparison.
+        cache_key = tuple(float(fn_args[name]) for name in self.source_param_names)
+        if self._cached_params is not None and cache_key == self._cached_params:
+            return self._cached_interpolator
+
+        try:
+            grid_sed = np.asarray(self.source.compute_sed(parameters=fn_args), dtype=float)
```
mcoughlin: Something like the below would ensure the same phases and wavelengths I think:

```
grid_sed = np.asarray(
    self.source.compute_sed(
        times=self.phases,
        wavelengths=self.wavelengths,
        parameters=fn_args,
    ),
    dtype=float,
)
```

jeremykubica:
@guillochon, thanks for adding this. Let me take a first look and then I can also add @mcoughlin.

src/lightcurvelynx/models/mosfit_models.py
```diff
@@ -0,0 +1,361 @@
...
+        Default: 100 points spanning [0, 200] days.
+    wavelengths : array_like, optional
+        The rest frame wavelength grid (in angstroms) on which to evaluate MOSFiT.
+        Default: 100 points spanning [1000, 25000] angstroms, which covers the Rubin
+        u through y range with room on either side for the caller's redshift.
+    time_extrapolation : FluxExtrapolationModel or tuple, optional
+        The extrapolation method to use for times outside the phase grid.
+        If nothing is provided, then the code adds zero padding.
+    source : object, optional
+        An already-constructed ``mosfit.lynx.LynxSource`` (or any object providing the
```
mcoughlin: If you agree with the suggestion below on compute_sed, the corresponding doc change might be:

```
    An already-constructed ``mosfit.lynx.LynxSource`` (or any object providing a
    ``compute_sed(times=..., wavelengths=..., parameters=...)`` interface) to use
    instead of building one. The source must honor the requested phase and wavelength
    grids and return an array with shape ``(len(phases), len(wavelengths))``.
    Mostly useful for testing and for sharing a single expensive MOSFiT model
    between several LightCurveLynx nodes.
```

mcoughlin (commented):
Really nice @guillochon !

I think it would be good to add a test where _ToyLynxSource.compute_sed() records times and wavelengths, then assert they equal TEST_PHASES and TEST_WAVES.

guillochon:
Thanks @mcoughlin! All three points addressed in c4dc5e3, and thanks @jeremykubica for picking up the reviewer assignment.

**1. Requesting the grid explicitly** — adopted as suggested:

```python
grid_sed = np.asarray(
    self.source.compute_sed(
        times=self.phases,
        wavelengths=self.wavelengths,
        parameters=fn_args,
    ),
    dtype=float,
)
```

I checked the cost before taking it, since `LynxSource` rebuilds MOSFiT's whole data-loading path on a grid change. It compares the grid it is handed against its own with `np.array_equal` and only rebuilds on a mismatch, so in the common case this is a no-op — instrumenting `LynxSource._build` gives one build across two evaluations with different parameters (first eval 2.2 s including the lazy build, second 0.00 s, SEDs different). It also turns out to fix a real hole: an externally supplied `source` built on a different grid now re-grids itself rather than silently returning an SED on the wrong one.

**2. Docstring for `source`** — taken verbatim, including the `(len(phases), len(wavelengths))` shape contract.

**3. Test recording the requested grids** — added `test_mosfit_wrapper_requests_its_own_grid`. `_ToyLynxSource` now records `last_times`/`last_wavelengths` and honors the grids it is handed, as `LynxSource` does. To make the assertion mean something I build the toy source on a *different* grid (11 phases over [0, 50], 5 waves over [2000, 8000]) and query at points matching neither, so the test fails if the wrapper ever forwards the query grid or falls back to the source's own:

```python
assert np.array_equal(source.last_times, TEST_PHASES)
assert np.array_equal(source.last_wavelengths, TEST_WAVES)
```

**One fix you did not ask for, but the review turned it up.** Writing the verification script above, I passed `pspin` and hit the wrapper's own error path, which told me the free parameters are `Pspin`, `Bfield`, `Mns`, `thetaPB`, `mejecta`, `kappa`, `kappagamma`, `vejecta`, `temperature`. MOSFiT's names are case sensitive, so the `pspin`/`bfield` in my class docstring and in the PR description were simply wrong and would not have worked. Both are corrected, and the docstring now says the names are MOSFiT's own and case sensitive.

13 tests pass, including the end-to-end one against MOSFiT from the `claude/lynx-sed-api` branch. `ruff` clean at the pinned v0.2.1. Still a draft pending guillochon/MOSFiT#252.

guillochon:
image

I wanted to make sure this worked properly end-to-end, so I took the fit from my 2014 paper fitting PS1-10jh (a tidal disruption event) and pretended it happened in a z = 0.5 galaxy, detected by LSST. Light curves are from this new MOSFiT module running in Lynx.

(Also if it isn't clear, the PR/code changes are using Claude. This is post is from me, the human.)

mcoughlin:
Looks good to me @guillochon !

src/lightcurvelynx/models/mosfit_models.py
```diff
@@ -0,0 +1,376 @@
...
+        ``LynxSource`` such as ``parameter_path``.
+
+    Note
+    ----
+    Building the underlying MOSFiT model is expensive, so it is deferred until the first
+    call to ``compute_sed`` and then reused. Because MOSFiT would rebuild that model for
+    every new grid, this wrapper never re-grids: it evaluates on ``phases`` and
+    ``wavelengths`` and linearly interpolates the result, returning zero outside them.
+
+    Building the model also copies MOSFiT's ``modules`` directory into the current working
```
jeremykubica: Can you elaborate on why this copy is needed by MOSFiT? From the MOSFiT code, it looks like the user should be able to access all the models and bandpasses through `mosfit.modules`. That should already be accessible if the user has pip installed mosfit into the current environment.

tests/lightcurvelynx/models/test_mosfit_models.py
```diff
@@ -0,0 +1,357 @@
...
+    """Test that LightCurveLynx, and not MOSFiT, applies the redshift.
+
+    MOSFiT's Lynx output is a rest frame SED at 10 pc, so the wrapper leaves
+    ``apply_redshift`` on and lets the superclass do the time dilation, the wavelength
+    shift and the distance dimming.
+    """
+    t0 = 64350.0
+    redshift = 1.0
+    model = MOSFiTWrapperModel(
+        "slsn",
```
jeremykubica: Given that this is using a real source, wouldn't it also do a force copy of the modules directory? And need the `monkeypatch.chdir(tmp_path)`

jeremykubica (commented):
The structure of the code and the wrapper look good. Two blockers on approving:
1) I would like to understand the copy of the `modules` directory and whether we can avoid that as could complicate some development workflows.
2) I would like to wait the final approval until the 2.1 is released, so we can run the github test suite using the full dependencies.

src/lightcurvelynx/models/mosfit_models.py
guillochon: Hi @jeremykubica , yes that's design holdover in MOSFiT. I've been meaning to change that behavior, and this is a good excuse to do that. I'll update the 2.1 MOSFiT PR then circle back to this one (might require no changes here).

guillochon: Should be fixed now upstream in the 2.1 PR, no more copying the modules directory into the working directory. Removed the note about the copy in this PR since it's no longer true.

jeremykubica: Thanks @guillochon!

jeremykubica:
This all looks really good. Thank you so much for putting this together. Since we need to wait on a merge for the 2.1 release of MOSFiT, I will hold off the final approval until we have that and can run the full tests.

guillochon:
Release 2.1 of MOSFiT has been cut and has been shipped to pypi. I have marked the PR ready for review.

jeremykubica:
Wow. That was fast. Can you uncomment the MOSFiT dependencies on the pyproject.toml so they get picked up by the test?

jeremykubica (approved):
This looks good. Thank you for adding it.

PR #978, opened 2026-09-22
Mdwarf

SHDaniels:
## Change Description

Adds M Dwarf Flare model that models m dwarf flares. For now I have also put the pzflow pkl file in the same directory, but we should move it somewhere else (open to suggestions). Will someone tell me how/where to make unit tests for this? (and in general the code could use another pair of eyes by someone who's familiar with lightcurvelynx.)

## Code Quality
- [x] I have read the Contribution Guide and agree to the Code of Conduct
- [x] My code follows the code style of this project
- [x] My code builds (or compiles) cleanly without any errors or warnings
- [x] My code contains relevant comments and necessary documentation

### DIFF
--- a/src/lightcurvelynx/models/mdwarf_flare_model.py
+++ b/src/lightcurvelynx/models/mdwarf_flare_model.py
@@ -1,0 +1,107 @@
+"""Models to simulate M-Dwarf Flares."""
+
+import numpy as np
+
+from lightcurvelynx.astro_utils.black_body import black_body_luminosity_density_per_solid
+from lightcurvelynx.models.physical_model import SEDModel
+
+
+class MDwarfFlareModel(SEDModel):
+    """An M-Dwarf Flare Model. The M-dwarf is modeled as the sum of several blackbody
+    functions (each of which has a temperature and a peak amplitude):
+
+    * The underlying star (amplitude is constant).
+    * The hot flare peak
+    * The cold flare peak
+
+    Parameterized values include:
+
+    * dec - The object's declination in degrees. [from BasePhysicalModel]
+    * distance - The object's luminosity distance in pc. [from BasePhysicalModel]
+    * ra - The object's right ascension in degrees. [from BasePhysicalModel]
+    * redshift - The object's redshift. [from BasePhysicalModel]
+    * t0 - The t0 of the zero phase, date. [from BasePhysicalModel. Not used.]
+
+    Parameters
+    ----------
+    star_temp : parameter, optional
+        The temperature of the star in Kelvins.
+    star_radius : parameter, optional
+        The radius of the star in cm.
+    hot_flare_temp : parameter, optional
+        The temperature of the hot flare in Kelvins.
+    cold_flare_temp : parameter, optional
+        The temperature of the cold flare in Kelvins.
+    flare_radius : parameter, optional
+        The radius of the combined hot and cold flare???
+    """
+
+    def __init__(
+        self,
+        *,
+        star_temp=None,
+        star_radius=None,
+        hot_flare_temp=None,
+        cold_flare_temp=None,
+        flare_radius=None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        # Add each model parameter, so it can be supported by the sampler.
+        self.add_parameter(
+            "star_temp", value=star_temp, description="The temperature of the star in Kelvins."
+        )
+        self.add_parameter("star_radius", value=star_radius, description="The radius of the star in cm.")
+        self.add_parameter(
+            "hot_flare_temp", value=hot_flare_temp, description="The temperature of the hot flare in Kelvins."
+        )
+        self.add_parameter(
+            "cold_flare_temp",
+            value=cold_flare_temp,
+            description="The temperature of the cold flare in Kelvins.",
+        )
+        self.add_parameter(
+            "flare_radius",
+            value=flare_radius,
+            description="The radius of the combined hot and cold flare in cm.",
+        )
+
+    def compute_sed(self, times, wavelengths, graph_state):
+        """Draw effect-free observer frame flux densities.
+
+        Parameters
+        ----------
+        times : numpy.ndarray
+            A length T array of observer frame timestamps in MJD.
+        wavelengths : numpy.ndarray, optional
+            A length N array of observer frame wavelengths (in angstroms).
+        graph_state : GraphState
+            An object mapping graph parameters to their values.
+
+        Returns
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of observer frame SED values (in nJy).
+        """
+        params = self.get_local_params(graph_state)
+        num_times = len(times)
+        num_waves = len(wavelengths)
+
+        flux_density = np.zeros((num_times, num_waves))
+
+        # If the star's temperature is defined, add it to the result flux.
+        if params["star_temp"] is not None:
+            flux_density += black_body_luminosity_density_per_solid(
+                wavelengths,
+                params["star_temp"],
+                params["star_radius"],
+            )
+
+        # Compute the size of the flare for each phase (time relative to t0).
+        # Compute and add in the contribution of the hot and cold flare
+        # using those blackbody curves.
+        # phase = times - params["t0"]
+
+        # Return the result.
+        return flux_density
--- a/tests/lightcurvelynx/models/test_mdwarf_flare_model.py
+++ b/tests/lightcurvelynx/models/test_mdwarf_flare_model.py
@@ -1,0 +1,3 @@
+def test_mdwarf_flare_model() -> None:
+    """Test that we can create and sample a MDwarfFlareModel object."""
+    pass


src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
+"""Models to simulate M-Dwarf Flares."""
+
+import numpy as np
+
+from lightcurvelynx.astro_utils.black_body import black_body_luminosity_density_per_solid
+from lightcurvelynx.models.physical_model import SEDModel
+from lightcurvelynx.math_nodes.np_random import NumpyRandomFunc
+from scipy import special
+from scipy.stats import binned_statistic
+from astropy.modeling.models import BlackBody #there may be an inbuilt way to do this - check astro_utils
```
jeremykubica: There is a function in `astro_utils`, but including this directly is fine if it is a better way to do things.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+    * t0 - The t0 of the zero phase, date. [from BasePhysicalModel]
+
+    Parameters
+    ----------
+    star_temp : parameter, optional
+        The temperature of the star in Kelvins.
+    star_radius : parameter, optional
+        The radius of the star in cm.
+    fwhm : parameter, optional
+        The log of the full width half max of the flare in time.
```
jeremykubica: Users confuse this with the FWHM for seeing? Maybe it should be `flare_fwhm`?

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        balmer_jump_ratio=None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        if star_temp is None or star_radius is None or fwhm is None or amplitude is None:
+            #must have put in all or none of the params
+            if star_temp is not None or star_radius is not None or fwhm is not None or amplitude is not None:
+                raise ValueError("must enter all or none of star_temp, star_radius, fwhm, and amplitude")
+            from pzflow import Flow #conditional install
```
jeremykubica: Since pzflow isn't installed by default, it would be better to wrap this in a try block that gives a more detailed error:
```
        try:
            from pzflow import Flow
        except ImportError as err:  # pragma: no cover
            raise ImportError(
                "pzflow package is not installed by default. You can install it with "
                "`pip install pzflow` or `conda install conda-forge::pzflow`."
            ) from err
```

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            flow = Flow(file='flare_flow.pzflow.pkl') #can do from PZFlowNode when we fix the version mismatch
```
jeremykubica: As discussed in chat, let's move this into data/models. You can get that directory using:

```
from lightcurvelynx import _LIGHTCURVELYNX_BASE_DATA_DIR
```
at the top and then

```
flow = Flow(file=_LIGHTCURVELYNX_BASE_DATA_DIR / flare_flow.pzflow.pkl)
```
Since `_LIGHTCURVELYNX_BASE_DATA_DIR` is a Path object, the slash will handle the correct path append.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            value=balmer_jump_ratio,
+            description="The ratio of the spectral intensity below the balmer jump to above the balmer jump (which is a blackbody at flare_temp)",
+        )
+        self.add_parameter(
+            "flare_temp",
+            value=flare_temp,
+            description="The temperature of the cold part of the flare in Kelvins."
+        )
+        if not self.has_valid_param("distance"):
+            sampler = MilkyWayCoordSampler(seed=42, node_label="mw") #change seed?
```
jeremykubica: We should leave the seed unspecified. If people want to control randomness, it is better for them to pass a random number generator into the overall sampling flow. That way they get consistent (controlled) numbers for everything.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            self.set_parameter("distance", value=sampler.distance_pc)
+            if self.has_valid_param("ra") or self.has_valid_param("dec"):
```
jeremykubica: Nice check!

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            if star_temp is not None or star_radius is not None or fwhm is not None or amplitude is not None:
+                raise ValueError("must enter all or none of star_temp, star_radius, fwhm, and amplitude")
+            from pzflow import Flow #conditional install
+            flow = Flow(file='flare_flow.pzflow.pkl') #can do from PZFlowNode when we fix the version mismatch
+            node = PZFlowNode(flow) 
+            #the node has them in log space
+            star_temp = BasicMathNode("10 ** log_teff", log_teff=node.logTeff)
+            star_radius = BasicMathNode("10 ** log_radius", log_radius=node.logRadius)
+            fwhm = BasicMathNode("10 ** log_fwhm", log_fwhm=node.logFWHM)
+            amplitude = BasicMathNode("10 ** log_amp", log_amp=node.logAmp)
```
jeremykubica: Nice!

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            self.set_parameter("ra", value=sampler.ra)
+            self.set_parameter("dec", value=sampler.dec)
+
+
+            
+        
+    
+
+    def _flare_eqn(self, t,tpeak,fwhm,amplitude):
+        '''
```
jeremykubica: Even though it is internal, the doc string here should include an Parameters section with a list of parameters (including units). It will help people trace through the code later.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        The equation that defines the shape for the Continuous Flare Model. taken from Tovar Mendoza et al. (2022) DOI 10.3847/1538-3881/ac6fe6
```
jeremykubica: You can do `@cite_function` (imported via `from citation_compass import cite_function`) and then add a "References" section to capture this. Like:

```
References
-----------
Tovar Mendoza et al. (2022) DOI 10.3847/1538-3881/ac6fe6
```

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+
+        f2 = 1-f1
+
+        eqn = ((1 / 2) * np.sqrt(np.pi) * A * C * f1 * np.exp(-D1 * t + ((B / C) + (D1 * C / 2)) ** 2)
+                            * special.erfc(((B - t) / C) + (C * D1 / 2))) + ((1 / 2) * np.sqrt(np.pi) * A * C * f2
+                            * np.exp(-D2 * t+ ((B / C) + (D2 * C / 2)) ** 2) * special.erfc(((B - t) / C) + (C * D2 / 2)))
+        return eqn * amplitude
+
+
+    def _flare_model(self, t,tpeak, fwhm, amplitude, upsample=False, uptime=10):
```
jeremykubica: You can also use `@cite_function` to pick up the references in this doc string.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+    
+        References
+        --------------
+        Tovar Mendoza et al. (2022) DOI 10.3847/1538-3881/ac6fe6
+        Davenport et al. (2014) http://arxiv.org/abs/1411.3723
+        Jackman et al. (2018) https://arxiv.org/abs/1804.03377
+    
+        Parameters
+        ----------
+        t : 1-d array
```
jeremykubica: I think this should be `t : numpy.ndarray`.

jeremykubica: The linter will complain about single letter variable names, so you might also want to call it time.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            The time array to evaluate the flare over
+    
+        tpeak : float
+            The center time of the flare peak
+    
+        fwhm : float
+            The Full Width at Half Maximum, timescale of the flare
```
jeremykubica: (in days)

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+    
+        amplitude : float
+            The amplitude of the flare
```
jeremykubica: relative to the star.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            The flux of the flare model evaluated at each time
+    
+            A continuous flare template whose shape is defined by the convolution of a Gaussian and double exponential
+            and can be parameterized by three parameters: center time (tpeak), FWHM, and ampitude
+        '''
+    
+        t_new = (t-tpeak)/fwhm
+    
+        if upsample:
+            dt = np.nanmedian(np.diff(np.abs(t_new)))
```
jeremykubica: Do we expect any of the times to be NaN? If not we could just use `np.median`. 

nanmedian is also fine. I was just curious as to the input.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        intensity = bb_low(wavelengths)
+        balmer_jump = 3645 * u.AA #angstroms
+        
+        
+        intensity[wavelengths < balmer_jump] *= balmer_jump_ratio
+        return intensity 
+        
+    def _tess_passband(self, wavelengths):
+        """Interpolate TESS T(lambda) onto arbitrary wavelengths. 0 outside range."""
+        #FIX?
```
jeremykubica: I'm not sure what the comment means. Is there something to fix?

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        Collapse a spectrum onto the TESS band: total_flux = integral of
+        spectrum(lambda) * T(lambda) dlambda / lambda .
+
+        Works on a plain 1D spectrum (n_wave,) or a stack with the wavelength
+        axis at `axis` (e.g. an (n_wave, n_time) array — pass axis=0, the
+        default).
+
+        Parameters
+        ----------
+        spectrum : Quantity, wavelength axis at `axis`. Should be in terms of frequency
```
jeremykubica: These should be formatted like the other parameters in this file:

variable name : type
    Description

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        wavelengths : Quantity, shape (n_wave,) — matches spectrum's wave axis
+        axis : which axis of `spectrum` is the wavelength axis
+
+        Returns
+        -------
+        Quantity, spectrum with the wavelength axis integrated out
```
jeremykubica: Should be:

```
Quantity
    The spectrum with the wavelength axis integrated out.
```

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        """
+        T_lambda = self._tess_passband(wavelengths)
+        wl = wavelengths .to(u.AA).value
```
jeremykubica: Extra space after `wavelengths`

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        
+        
+        intensity[wavelengths < balmer_jump] *= balmer_jump_ratio
+        return intensity 
+        
+    def _tess_passband(self, wavelengths):
+        """Interpolate TESS T(lambda) onto arbitrary wavelengths. 0 outside range."""
+        #FIX?
+        _filt = SvoFps.get_transmission_data('TESS/TESS.Red')
+        TESS_WAVE = np.asarray(_filt['Wavelength']) * u.AA      # SVO gives this in Angstrom
```
jeremykubica: All local variable names should be lowercase.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        -------
+        Quantity, spectrum with the wavelength axis integrated out
+        """
+        T_lambda = self._tess_passband(wavelengths)
+        wl = wavelengths .to(u.AA).value
+        w = T_lambda * wl
+        shape = [1] * spectrum.ndim
+        shape[axis] = -1
+        w = w.reshape(shape)
+        # unit = spectrum.unit * u.AA #not sure if we need this?
```
jeremykubica: I don't think that unit is needed, so you can remove that commented line.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        Returns
+        -------
+        Quantity, shape (n_wave,), units erg/(Hz s sr cm^2) 
+        it is now in terms of frequency and not wavelength
+        """
+        if not isinstance(temp_star, u.Quantity):
+            temp_star = temp_star * u.K
+        bb = BlackBody(temperature=temp_star)
+        I_lam = bb(wavelengths) 
+        return  I_lam #earlier version had np.pi * u.sr * radius_star**2 *
```
jeremykubica: It looks like the comment is cut off.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+        params = self.get_local_params(graph_state) #should have everything that was instantiated
+        num_times = len(times)
+        num_waves = len(wavelengths)
+        if not isinstance(wavelengths, u.Quantity):
+            wavelengths = wavelengths * u.AA #angstroms
+        
+        flux_density = np.zeros((num_times, num_waves))
+
+        constants = np.pi  * params["star_radius"]**2 * u.sr 
+        if not isinstance(params["star_radius"], u.Quantity):
```
jeremykubica: The parameters that come out of `get_local_params` will always be plain floats (not astropy units). So you can just do:
```
constants = np.pi  * params["star_radius"]**2 * u.sr * (u.cm**2)
```

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+            constants = constants * (u.cm**2) 
+            #we are losing the star radius units if we don't do this
+
+        print('constants units', constants.unit)
```
jeremykubica: You can remove this after applying the units above.

src/lightcurvelynx/models/mdwarf_flare_model.py
```diff
@@ -0,0 +1,356 @@
...
+
+        Returns
+        -------
+        flux_density : numpy.ndarray
+            A length T x N matrix of observer frame SED values (in nJy).
+        """
+        params = self.get_local_params(graph_state) #should have everything that was instantiated
+        num_times = len(times)
+        num_waves = len(wavelengths)
+        if not isinstance(wavelengths, u.Quantity):
```
jeremykubica: The wavelength parameter will always come into the function without astropy units attached. So you don't need to check. You can just do:
```
wavelengths = wavelengths * u.AA
```

jeremykubica (commented):
High level comments:
1) You will want to run a linter to clean up all the small formatting issues that the precommit is picking up. You can do this by installing `ruff` in your current environment (`pip install ruff`) then running it (`ruff format`) from the lightcurvelynx directory. This will auto-fix a lot of the formatting.
2) You should add a test in `tests/lightcurvelynx/models/test_mdwarf_flare_model.py‎`. I would recommend basing it one of the other test files in that directory. It should cover creating and querying the MDwarfFlareModel object.

src/lightcurvelynx/models/mdwarf_flare_model.py
SHDaniels: changing it to flare_fwhm and amplitude to flare_amplitude to match

src/lightcurvelynx/models/mdwarf_flare_model.py
SHDaniels: honestly i have not ever used the upsample functionality but I left it in there since it's in the original Mendoza code. I think we would have other problems if any of our times or other parameter values were NaNs, though

src/lightcurvelynx/models/mdwarf_flare_model.py
SHDaniels: clearly i thought there was at some point but it all seems fine now

SHDaniels: however I think lightcurvelynx might have some functionality that duplicates this? so maybe that was the thing to fix

